Issue #25, opened 2021-02-03
SecondaryDeserializer should not be a singleton


FabianF92:
Hi,

we are migrating our deserializers to GLUE using the fallback option with `SecondaryDeserializer`.
We have both Avro `SpecificDatumReader` and `GenericDatumReader` in our project so we need to have different types of `SecondaryDeserializer`. 
However, the `SecondaryDeserializer` is a singleton and we can't have two instances of it.
It would be better if the we can have one `SecondaryDeserializer` instance per `AWSKafkaAvroDeserializer`.

mohitpali:
I can update the changelog on your behalf. Thanks for the PR.

PR #26, opened 2021-02-04
One SecondaryDeserializer instance per AWSKafkaAvroDeserializer

lagoonos:
*Issue #25 *

*Description of changes:*

Refactored AWSKafkaAvroDeserializer to have it's own SecondaryDeserilizer instance.


mohitpali (changes requested):
Thanks for sending this. Would you be able to add a simple unit test for this case ?

lagoonos:
> Thanks for sending this. Would you be able to add a simple unit test for this case ?

Hi I added a unit test to assure two instances of the _AWSKafkaAvroDeserializer_ can have indenependent deserializers (with independent configurations)

mohitpali (approved):

PR #40, opened 2021-04-01
Upgrading to 1.0.2

blacktooth:
* Update to latest secure version of Guava due to security issue.


mohitpali (changes requested):
The upgrade would require the version to be updated in all the poms to 1.0.2 as well.

mohitpali:
Could you also please rebase against master and then force push to your fork and create review so that the previous commits don't show up in this request.

Issue #33, opened 2021-03-01
Extend customization of schema naming strategy


jagpat:
Hello,

We would like to publish schema from kafka connect to glue schema registry. We would like to name the created schemas in glue registry as \<topicname\>-key and \<topicname\>-value. 
So we implemented AWSSchemaNamingStrategy interface using CustomerSchemaNamingStrategy as an example. However the overridden method "public String getSchemaName(String transportName, Object data)" do not provide information on whether the schema is for key or for value so we cannot append "-key" or "-value" to the topicname to derive the schema name. Since the "AWSKafkaAvroSerializer::configure()" method already has isKey as the parameter, would you be able to extend the AWSSchemaNamingStrategy to include a method  getSchemaName(String transportName, Object data, boolean isKey)" ? Or is there any other way we would be able to determine whether the schema name is for the key or value ? Thanks!

mohitpali:
Thanks for reaching out, we will look into this.

PR #42, opened 2021-04-05
Update schema naming strategy to allow key and value based namings

mohitpali:
*Issue #, https://github.com/awslabs/aws-glue-schema-registry/issues/33*

*Description of changes: Extend customization of schema naming strategy by key and value*

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -213,6 +213,7 @@
                         <!-- Ignore exception classes -->
                         <exclude>**/AWSIncompatibleDataException*</exclude>
                         <exclude>**/AWSSchemaRegistryException*</exclude>
+                        <exclude>**/AWSSchemaNamingStrategy*</exclude>
                     </excludes>
                 </configuration>
                 <executions>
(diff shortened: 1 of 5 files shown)


common/pom.xml
```diff
@@ -213,6 +213,7 @@
                         <!-- Ignore exception classes -->
                         <exclude>**/AWSIncompatibleDataException*</exclude>
                         <exclude>**/AWSSchemaRegistryException*</exclude>
+                        <exclude>**/AWSSchemaNamingStrategy*</exclude>
```
blacktooth, line 216: Why do we have to exclude this?

mohitpali: Because it is in an interface and by default we don't call default methods.

blacktooth: Can we test it by creating an implementation that doesn't call it?

blacktooth (approved):

common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaNamingStrategy.java
```diff
@@ -20,9 +20,17 @@
...
-    default String getSchemaName(String transportName, Object data) {
+    default String getSchemaName(String transportName,
+                                 Object data) {
         return getSchemaName(transportName);
     }
+
+    default String getSchemaName(String transportName,
+                                 Object data,
+                                 boolean isKey) {
+        return isKey ? getSchemaName(transportName) : getSchemaName(transportName, data);
```
OneCricketeer, line 31: So, by default `data` is ignored? I assume you did this for backwards compatibility, but it leads to issues discussed in #93

Issue #44, opened 2021-05-11
Serializer / Deserializer methods should rewind the buffers


blacktooth:
Methods like [canDeserialize](https://github.com/awslabs/aws-glue-schema-registry/blob/1bfc2cfb8a5c108964fadf7777e3d29b38e7ea9b/avro-serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/AWSDeserializer.java#L174) and [getData](https://github.com/awslabs/aws-glue-schema-registry/blob/1bfc2cfb8a5c108964fadf7777e3d29b38e7ea9b/avro-serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/AWSDeserializer.java#L116) should return rewinded buffer so that clients don't have to rewind them themselves.

PR #45, opened 2021-05-18
Fixes bug that leaves the ByteBuffer at non-starting position.

blacktooth:
*Issue #, if available:* https://github.com/awslabs/aws-glue-schema-registry/issues/44

*Description of changes:*

Fixing the bug that leaves the buffer at a non-starting position. In case of `canDeserialize` method, clients have to rewind the buffer by themselves for non-Schema Registry messages.

### DIFF
--- a/avro-serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/AWSDeserializerDataParser.java
+++ b/avro-serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/AWSDeserializerDataParser.java
@@ -64,19 +64,20 @@ public final class AWSDeserializerDataParser {
      *                                      schema registry
      */
     public UUID getSchemaVersionId(ByteBuffer byteBuffer) {
-        //Make sure we have valid byteBuffer.
-        validateData(byteBuffer);
-
-        // Ensure that we rewind to the start to avoid side-effects on byteBuffer
-        // instance
         byteBuffer.rewind();
-        // Skip HEADER_VERSION_BYTE
-        byteBuffer.get();
-        // Skip COMPRESSION_BYTE
-        byteBuffer.get();
+        // Ensure that we are not changing the buffer position.
+        ByteBuffer slicedBuffer = byteBuffer.slice();
 
-        long mostSigBits = byteBuffer.getLong();
-        long leastSigBits = byteBuffer.getLong();
+        //Make sure we have valid byteBuffer.
+        validateData(slicedBuffer);
+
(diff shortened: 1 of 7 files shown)


avro-serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/AWSDeserializerDataParser.java
```diff
@@ -64,19 +64,20 @@ public static AWSDeserializerDataParser getInstance() {
      *                                      schema registry
      */
     public UUID getSchemaVersionId(ByteBuffer byteBuffer) {
+        byteBuffer.rewind();
```
m4braham, line 67: Since you now slice the input buffer, why rewind?  I reckon you should leave the buffer without any side effects.

Actually, I’m not even sure this is the right approach.  I would argue that it should be the responsibility of the deserializer’s client, to handle buffer positions.  This is the KCL.  It ought to slice the buffer and use the slice to test for schema encoding.  If this test is positive then it should decode the schema, using another slice.  It is somewhat unfortunate that the deserializer interface has access to the “plain” data through a third method but so then the KCL ought to call that method, with a third slice.

So I don’t think that the deserializer (or this class) should be responsible for maintaining buffer positions…

m4braham (changes requested):
See my other comment, I believe that handling buffer positions ought to be a responsibility of the KCL code that drives the deserializer…

avro-serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/AWSDeserializerDataParser.java
blacktooth: Thanks for reviewing!

>Since you now slice the input buffer, why rewind? I reckon you should leave the buffer without any side effects.

When a buffer with non-initial position is sliced, the sliced buffer starts from non-initial position and previous bytes are not accessible. Clients who might used to sending buffer with non-rewinded state assuming the GSR library will rewind might break, so kept it same. 

KCL is one of the consumers of the library, customers can directly use this library. We want this behavior consistent across all the clients so that every doesn't have to make this change. It's also good to hide the complexity from clients. Do you see any side-effects with this approach?

m4braham: I understand your concern regarding existing clients that might break if the code doesn't rewind the buffer... I hadn't considered this.

PR #58, opened 2021-07-09
Add additional examples of configuring Kafka Connect and clarification on what property names are expected.

### LINKED ISSUE #56: Need to be able to specify a GSR
By default, schemas are getting created in default-registry. Is there a way to specify a different registry in GSR?

dude0001:
*Issue #, if available:* 
Fixes #56 

*Description of changes:*

It should be obvious by looking at the code what properties are expected when configuring this in Kafka Connect, but this documentation update should give users a nudge in the right direction where to look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,8 @@ mvn dependency:copy-dependencies
 
 * Configure Kafka Connectors with following properties
 
+When configuring Kafka Connect workers or connectors, use the value of the string constant properties in the [AWSSchemaRegistryConstants](https://github.com/awslabs/aws-glue-schema-registry/blob/master/common/src/main/java/com/amazonaws/services/schemaregistry/utils/AWSSchemaRegistryConstants.java#L20) class to configure the AWSKafkaAvroConverter.
+
 ```java
     key.converter=com.amazonaws.services.schemaregistry.kafkaconnect.AWSKafkaAvroConverter
     value.converter=com.amazonaws.services.schemaregistry.kafkaconnect.AWSKafkaAvroConverter
@@ -388,6 +390,14 @@ mvn dependency:copy-dependencies
     key.converter.avroRecordType=GENERIC_RECORD
     value.converter.avroRecordType=GENERIC_RECORD
 ```
+
+As Glue Schema Registry is a fully managed service by AWS, there is no notion of schema registry URLs. Name of the registry (within the same AWS account) can be optionally configured using following options. If not specified, default-registry is used.
+
+```java
+    key.converter.registry.name=my-registry
+    value.converter.registry.name=my-registry
+```
+
 * Add command below to *Launch mode* section under *kafka-run-class.sh*
 
 ``` 


README.md
```diff
@@ -378,11 +378,21 @@ mvn dependency:copy-dependencies
 
 * Configure Kafka Connectors with following properties
 
+When configuring the worker or connectors in Kafka Connect, refer to the string constant properties of the [AWSSchemaRegistryConstants](https://github.com/awslabs/aws-glue-schema-registry/blob/master/common/src/main/java/com/amazonaws/services/schemaregistry/utils/AWSSchemaRegistryConstants.java#L20) class.
+
+Examples:
+
+AWSSchemaRegistryConstants.AWS_REGION = "region"
```
blacktooth, line 385: IMO, this might confuse users as it's followed by code. We don't need the Examples as the link above points to code.

README.md
```diff
@@ -378,11 +378,21 @@ mvn dependency:copy-dependencies
...
+AWSSchemaRegistryConstants.REGISTRY_NAME = "registry.name"
+AWSSchemaRegistryConstants.COMPATIBILITY_SETTING = "compatibility";
+
 ```java
     key.converter=com.amazonaws.services.schemaregistry.kafkaconnect.AWSKafkaAvroConverter
     value.converter=com.amazonaws.services.schemaregistry.kafkaconnect.AWSKafkaAvroConverter
     key.converter.region=ca-central-1
+    key.converter.registry.name=my-registry
```
blacktooth, line 393: Can you please add a comment above this line to indicate that this is optional? We want users to get started with minimal possible configuration.

blacktooth:
Thanks for submitting this improvement!

README.md
dude0001: Good call, that makes sense to me. The examples have been removed.

README.md
dude0001: The registry name examples were copied from some of the other configuration examples above in the file. I was not aware of the intent to keep this as an example of the minimal possible configuration. However, if that is the goal here seems to be inconsistency. This led me to assume specifying a specific GSR resource/endpoint wasn't implemented for Kafka Connect. I believe this was because a) I couldn't get it to work with the property names I used (that I mostly copied from the other examples). b) you usually need a URL to the schema registry in Kafka Connect, so it isn't obvious what to do here. In that sense, I added an additional note about this scenario. Maybe that makes more sense.

Please take another look. Also, let me know if I should squash.

README.md
```diff
@@ -388,6 +390,14 @@ mvn dependency:copy-dependencies
     key.converter.avroRecordType=GENERIC_RECORD
     value.converter.avroRecordType=GENERIC_RECORD
 ```
+
+Kafka Connect schema registry convertes are typically configured with a URL to the desired schema registery endpoint. GSR does not have the notion of a URL endpoint. The service is instead accessed with AWS API calls specifying the name of the GSR resource. You can optionally specify the name of the desired GSR resource. If registry name isn't provided a default GSR resource is used.
```
blacktooth, line 394: We can shorten this a bit,

"As Glue Schema Registry is a fully managed service by AWS, there is no notion of schema registry URLs. Name of the registry (within the same AWS account) can be optionally configured using following options. If not specified, `default-registry` is used."

README.md
blacktooth: It would be great if you can squash the commits.

README.md
dude0001: That is concise and gets the point across. Thank you for the suggestion, I've updated my PR with the change.

README.md
dude0001: Done

dude0001:
I believe this is ready to be merged.

blacktooth (approved):

blacktooth:
Thanks for improving our documentation! Merging now!

Issue #63, opened 2021-07-21
Too much logging for AWSKafkaAvroSerializer (part two)


snemarch:
Hi, we're getting a "Schema Version Id is null. Trying to register the schema" log for every message serialized with the AWSKafkaAvroSerializer.

It seems that this was changed from info to debug in #9 , but changed back to info as part of #49 (commit a7017fb7a418f30d6302d3822e3464388228a72c). Re-changing the log level to debug would remove the spam, but I'm contemplating if there could be some underlying issues?

I haven't fully grokked the code, but in [AWSKafkaAvroSerializer#serialize](https://github.com/awslabs/aws-glue-schema-registry/blob/c9b9215fa5b976f27e251c0d38d18ae956c62f78/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/avro/AWSKafkaAvroSerializer.java#L102-L117) the null-check is against the final `schemaVersionId`... is there any feasible way to get this configured/initialized to be non-null?

My hunch is that either there's a bit of an architectural issue, or there's some configuration issue we've totally missed – it doesn't seem right that the lookup fails for each and every message. Fortunately the caching in `GlueSchemaRegistrySerializationFacade` seems to be working, though! 😅

blacktooth:
Hi,

Is the code able to register the schema version successfully? Do you see any exceptions? Does the schema version show. up in the Glue Schema Registry console?

blacktooth:
Re-directing the discussion to #64

blacktooth:
Thanks for reporting this! Looks like the code got reverted in this file. We will investigate and fix it.

snemarch:
> Thanks for reporting this! Looks like the code got reverted in this file. We will investigate and fix it.

Note that I think there's some underlying issue (possibly with our setup, we might very well have missed some configuration!), which will be masked out by re-changing the log level from info to debug.

Some more context, in case it helps: we're running a CDC pipeline: Postgres -> Debezium -> Kafka -> ElasticSearch, orchestrated with Kafka Connect. We're using AWS MSK and Glue Registry, while Connect is running in our own Kubernetes cluster.

PR #64, opened 2021-07-21
Reduce logging

tveon:
Fixes issue #63 

*Description of changes:*

- Add basic `.gitignore`
- Reduce logging level to debug for statements which are likely to occur frequently in production systems

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,7 @@
+#Ignore IntelliJ files
+.idea/
+*.iml
+
+#Ignore Maven files
+**/target
+
--- a/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/GlueSchemaRegistryKafkaSerializer.java
+++ b/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/GlueSchemaRegistryKafkaSerializer.java
@@ -98,7 +98,7 @@ public class GlueSchemaRegistryKafkaSerializer implements Serializer<Object> {
 
         UUID schemaVersionIdFromRegistry = null;
         if (this.schemaVersionId == null) {
-            log.info("Schema Version Id is null. Trying to register the schema.");
+            log.debug("Schema Version Id is null. Trying to register the schema.");
             schemaVersionIdFromRegistry =
                     glueSchemaRegistrySerializationFacade.getOrRegisterSchemaVersion(prepareInput(data, topic, isKey));
         } else {
@@ -106,7 +106,7 @@ public class GlueSchemaRegistryKafkaSerializer implements Serializer<Object> {
         }
 
         if (schemaVersionIdFromRegistry != null) {
-            log.info("Schema Version Id received from the from schema registry: {}", schemaVersionIdFromRegistry);
+            log.debug("Schema Version Id received from the from schema registry: {}", schemaVersionIdFromRegistry);
             result = glueSchemaRegistrySerializationFacade.serialize(DataFormat.fromValue(dataFormat), data,
                                                                      schemaVersionIdFromRegistry);
         }
--- a/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/avro/AWSKafkaAvroSerializer.java
+++ b/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/avro/AWSKafkaAvroSerializer.java
@@ -74,7 +74,7 @@ public class AWSKafkaAvroSerializer implements Serializer<Object> {
     @Override
     public void configure(@NonNull Map<String, ?> configs,
                           boolean isKey) {
-        log.info("Configuring Amazon Schema Registry Service using these properties: {}", configs);
+        log.info("Configuring Amazon Glue Schema Registry Service using these properties: {}", configs);
         schemaName = GlueSchemaRegistryUtils.getInstance()
                 .getSchemaName(configs);
         this.isKey = isKey;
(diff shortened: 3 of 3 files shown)


blacktooth:
Are you facing performance issues due to these logging statements? These statements are logged during auto-registration flow which we don't expect to be a frequent operation. Is it different in your use-case?

snemarch:
> Are you facing performance issues due to these logging statements? These statements are logged during auto-registration flow which we don't expect to be a frequent operation. Is it different in your use-case?

See #63 🙂 – we get the log statement for every sent. Not sure of the performance implications, but drowns useful log messages and does have some log storage cost.

common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaRegistryClient.java
```diff
@@ -284,7 +284,7 @@ public UUID createSchema(String schemaName,
                              Map<String, String> metadata) throws AWSSchemaRegistryException {
         UUID schemaVersionId = null;
         try {
-            log.info("Auto Creating schema with schemaName: {} and schemaDefinition : {}", schemaName,
```
blacktooth, line 287: Are you seeing this logged for every request?

snemarch: We're only seeing the "Auto Creating schema" message once, and then the "Schema Version Id is null" on every message.

mohitpali: Thanks for fixing this. Would you mind updating it in class [GlueSchemaRegistryKafkaSerializer.java](https://github.com/awslabs/aws-glue-schema-registry/blob/master/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/GlueSchemaRegistryKafkaSerializer.java) as well. This class is a generic class that can be used with additional property called "dataFormat" and passing "JSON" or "AVRO".

snemarch: Note that re the information in #63, I think there might be an underlying issue, and changing the log level will just mask that.

mohitpali: I think this got lost in merge conflicts resolution. This has been a miss on our side. The schema auto creation should still be info because we don't expect that to happen very often. The Over logging issue is legit. Are you suggesting that there is an underlying issue with the library or your set-up particularly ? Either ways, we would be happy to help.

snemarch: Well, we're getting the "_Schema Version Id is null. Trying to register the schema_" log for every serialized message (which is fun when doing initial dump of a 200+ million row table 😀) – I'm *assuming* the schema identifier ought to be saved after (auto-)creation, but I don't know if it's an issue with the serializer, our configuration, or the interplay between Debezium and Kafka Connect.

As mentioned in #63, the caching in `GlueSchemaRegistrySerializationFacade` seems to be working, so we don't end up swamping Glue SR with requests – but my "_I haven't studied the code enough to understand it well_" hunch is that it probably shouldn't be the cache layer preventing this?

mohitpali: That logging message should be debug. Some customers have just one schemaVersionId and that can be hard coded in the AWSKafkaAvroSerializer for efficiency. Schema Identifier for your case would be saved in a in-memory cache later in the code path and the message you are seeing may confuse you that schema version id is null.

tveon: I _think_ @snemarch means, that there is no way to specify the `schemaVersionId` when using Kafka Connect - or documentation around when that would be applicable. 🙂

tveon: I've updated `GlueSchemaRegistryKafkaSerializer` to also use `debug` for the logging is `serialize`.

snemarch: @mohitpali - so, if I understand correctly:

1. Usecases where dynamic schema registration is used isn't meant to store the registered schemaVersionId, and the cache a crucial part of the design, not "just" an optimization.
2. Usecases where you can manually evolve schemas can have the schemaVersionId predefined.

Even with re-changing log level to DEBUG, would it be worth adding a comment to the codebase that it's normal to get this on a per-message basis for dynamic scenarios, and that caching further down ensures this is not a performance problem?

Thanks for your help in clearing up matters! :)

mohitpali: Good point, we can add these comments, thanks !

mohitpali: If you could please change the below line to info, we should be good to merge.
``
log.debug("Auto Creating schema with schemaName: {} and schemaDefinition : {}", schemaName, schemaDefinition);
``

snemarch: > If you could please change the below line to info, we should be good to merge.

@tveon is on vacation for the next few weeks, but I'll see if I can persuade him to do that one-line update :)

tveon: Done. 👍 🌴

mohitpali (approved):

snemarch:
Hello, good people, do you have any plans for when the next release is going to be? 😃

PS: I just noticed that while there's a 1.1.1 release on Maven Central, this GitHub repo only has tags for 1.0.2 and 1.1.0.

Issue #46, opened 2021-05-25
Support environment variables for Kafka Connect config


oscarkaypee:
Hi all,

We're using the latest Glue Schema Registry source with our dockerized Kafka Connect set up which uses environment variables for it's configuration https://docs.confluent.io/platform/current/installation/docker/config-reference.html#kconnect-long-configuration

This in theory means properties such as `value.converter.avroRecordType=GENERIC_RECORD` would be represented as `CONNECT_VALUE_CONVERTER_AVRORECORDTYPE: GENERIC_RECORD` in our environment variables for the container, following the Confluent docs above. However, this runs into: 
`[2021-05-24 12:00:31,125] INFO avroRecordType key is not present in the configs {avrorecordtype=GENERIC_RECORD, schemaautoregistrationenabled=true, region=eu-west-1} (com.amazonaws.services.schemaregistry.common.configs.GlueSchemaRegistryConfiguration)`

I've tried:
* `CONNECT_VALUE_CONVERTER_avroRecordType`
* `CONNECT_VALUE_CONVERTER_AVRORECORDTYPE`
* `CONNECT_VALUE_CONVERTER_AVRO_RECORD_TYPE`
* `CONNECT_VALUE_CONVERTER_AVRO__RECORD__TYPE`
* `CONNECT_VALUE_CONVERTER_AVRO___RECORD___TYPE`

Which all result in the same lowercase `avrorecordtype` in the configs e.g. `{avrorecordtype=GENERIC_RECORD}`

Is it possible to support setting these properties via environment variables? Perhaps by ignoring case in the presence check?

LinyuYao1021:
Hi,

Thanks for reaching out. Our configuration parameters are case sensitive. We will look deeper to see if we can support for insensitive ones.

WhiskyTangoFawks:
I would like to bump this issue. I ran into the same problem- the confluent docker container creates the configs with all lower case, so it's impossible to use this library with them without hacking a workaround.

mohitpali:
While i am looking into this, would you mind describing your use case ?

WhiskyTangoFawks:
[...]

PR #71, opened 2021-08-17
Make variables consistent

mohitpali:
*Issue #, if available:* https://github.com/awslabs/aws-glue-schema-registry/issues/46

*Description of changes:* Modified constant variable names to keep them lowercase and with Kafka convention to support use cases like connect.


integration-tests/kafka-connect-avro-configs/file-sink-standalone.properties
```diff
@@ -13,4 +13,4 @@ internal.value.converter.schemas.enable=true
 # with the converter we want to apply it to
 value.converter.endpoint=https://glue-gamma.us-east-1.amazonaws.com
 value.converter.region=us-east-1
-value.converter.avroRecordType=GENERIC_RECORD
\ No newline at end of file
+value.converter.avro.record.type=GENERIC_RECORD
```
blacktooth, line 16: Does this break configs of existing users?

mohitpali: The Constant names are not changing. There is a risk of customers are using it with the name directly rather than using constant names. But this will be a version bump, so existing customers will not break.

blacktooth:
We are putting this on hold to explore other backward compatible changes.

PR #77, opened 2021-09-13
Modify UserAgent to emit usage metrics

blacktooth:
*Description of changes:*

Appending a string to the UserAgent to track the usage of different features.

*Testing*

Sample UserAgent contains following string,

```
autoreg/0:compress/0:secdeser/0:app/kafka/1.1.2

autoreg/1:compress/1:secdeser/1:app/kafkastreams/1.1.2
```

### DIFF
--- a/common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaRegistryClient.java
+++ b/common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaRegistryClient.java
@@ -18,10 +18,18 @@ package com.amazonaws.services.schemaregistry.common;
 import com.amazonaws.services.schemaregistry.common.configs.GlueSchemaRegistryConfiguration;
 import com.amazonaws.services.schemaregistry.exception.AWSSchemaRegistryException;
 import com.amazonaws.services.schemaregistry.utils.AWSSchemaRegistryConstants;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.awscore.AwsRequestOverrideConfiguration;
+import software.amazon.awssdk.core.ApiName;
+import software.amazon.awssdk.core.SdkRequest;
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
+import software.amazon.awssdk.core.interceptor.Context;
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
 import software.amazon.awssdk.core.retry.RetryPolicy;
 import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
 import software.amazon.awssdk.regions.Region;
@@ -37,6 +45,7 @@ import software.amazon.awssdk.services.glue.model.GetSchemaVersionRequest;
 import software.amazon.awssdk.services.glue.model.GetSchemaVersionResponse;
 import software.amazon.awssdk.services.glue.model.GetTagsRequest;
 import software.amazon.awssdk.services.glue.model.GetTagsResponse;
+import software.amazon.awssdk.services.glue.model.GlueRequest;
 import software.amazon.awssdk.services.glue.model.MetadataKeyValuePair;
 import software.amazon.awssdk.services.glue.model.PutSchemaVersionMetadataRequest;
 import software.amazon.awssdk.services.glue.model.PutSchemaVersionMetadataResponse;
@@ -50,6 +59,7 @@ import software.amazon.awssdk.services.glue.model.SchemaId;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Map;
+import java.util.StringJoiner;
 import java.util.UUID;
 
 /**
@@ -74,11 +84,11 @@ public class AWSSchemaRegistryClient {
     public AWSSchemaRegistryClient(@NonNull AwsCredentialsProvider credentialsProvider,
                                    @NonNull GlueSchemaRegistryConfiguration glueSchemaRegistryConfiguration,
                                    @NonNull RetryPolicy retryPolicy) {
+        this.glueSchemaRegistryConfiguration = glueSchemaRegistryConfiguration;
         ClientOverrideConfiguration overrideConfiguration = ClientOverrideConfiguration.builder()
                 .retryPolicy(retryPolicy)
+                .addExecutionInterceptor(new UserAgentRequestInterceptor())
                 .build();
-
-        this.glueSchemaRegistryConfiguration = glueSchemaRegistryConfiguration;
 
         GlueClientBuilder glueClientBuilder = GlueClient
                 .builder()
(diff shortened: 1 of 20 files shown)


common/src/main/java/com/amazonaws/services/schemaregistry/utils/AWSSchemaRegistryConstants.java
```diff
@@ -158,6 +158,12 @@
      */
     public static final String JACKSON_DESERIALIZATION_FEATURES = "jacksonDeserializationFeatures";
 
+    /**
+     * Name of the UserAgent application using the serializer/de-serializer library.
+     * Ex: KPL, Kafka, KafkaConnect etc.
+     */
+    public static final String USER_AGENT_APP = "userAgentApp";
```
mohitpali, line 165: I would recommend naming the variables in kafka friendly format since we support kafka systems.
Change it to "user.agent.app"

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/UserAgents.java
```diff
@@ -0,0 +1,8 @@
+package com.amazonaws.services.schemaregistry.common.configs;
+
+public class UserAgents {
+    public static final String KAFKA = "kafka";
+    public static final String KAFKASTREAMS = "kafkastreams";
+    public static final String KAFKACONNECT = "kafkaconnect";
+    public static final String FLINK = "flink";
```
mohitpali, line 7: Kinesis ?

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
```diff
@@ -87,13 +94,34 @@ private void buildSchemaRegistryConfigs(Map<String, ?> configs) {
...
     }
 
     private void buildCacheConfigs(Map<String, ?> configs) {
         validateAndSetCacheSize(configs);
         validateAndSetCacheTTL(configs);
     }
 
+    private void validateAndSetUserAgent(Map<String, ?> configs) {
+        if (isPresent(configs, AWSSchemaRegistryConstants.USER_AGENT_APP)) {
+            this.userAgentApp = (String) configs.get(AWSSchemaRegistryConstants.USER_AGENT_APP);
```
mohitpali, line 121: ``String.valueOf`` is better to avoid any casting exceptions.

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
```diff
@@ -87,13 +94,34 @@ private void buildSchemaRegistryConfigs(Map<String, ?> configs) {
...
+        }
     }
 
     private void buildCacheConfigs(Map<String, ?> configs) {
         validateAndSetCacheSize(configs);
         validateAndSetCacheTTL(configs);
     }
 
+    private void validateAndSetUserAgent(Map<String, ?> configs) {
+        if (isPresent(configs, AWSSchemaRegistryConstants.USER_AGENT_APP)) {
```
mohitpali, line 120: What if customer overrides this configuration, how do we prevent customer from passing junk value or a custom user agent ?

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/UserAgents.java
blacktooth: KCL and KPL will be passed from the respective libraries. I didn't want those repos to depend on this class for simple constants.

common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaRegistryClient.java
```diff
@@ -554,4 +564,58 @@ public GetTagsResponse querySchemaTags(String schemaDefinition, String schemaNam
...
+        private ApiName getApiName() {
+            return ApiName.builder()
+                .version(com.amazonaws.services.schemaregistry.common.MavenPackaging.VERSION)
+                .name(buildUserAgentSuffix())
+                .build();
+        }
+
+        private String buildUserAgentSuffix() {
+            Map<String, String> userAgentSuffixItems = ImmutableMap.of(
+                "autoreg", glueSchemaRegistryConfiguration.isSchemaAutoRegistrationEnabled() ? ONE : ZERO,
```
mohitpali, line 606: nit : any specific reason or limit that they have to be short and small case ?

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
blacktooth: > `String.valueOf` is better to avoid any casting exceptions.

I think it's good to fail this during validation if the UserAgentApp value is not set as String to avoid side-effects like setting this to Object hash code.

serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/GlueSchemaRegistryDeserializationFacade.java
```diff
@@ -97,6 +97,15 @@ public GlueSchemaRegistryDeserializationFacade(@NonNull GlueSchemaRegistryConfig
         this.cache = GlueSchemaRegistryDeserializerCache.getInstance(glueSchemaRegistryConfiguration);
     }
 
+    /**
+     * Method to override user-agent app name for the de-serializer.
+     * This overrides the previously set value in GlueSchemaRegistryConfiguration.
+     * @param name AppName
+     */
+    public void overrideUserAgentApp(String name) {
+        this.glueSchemaRegistryConfiguration.setUserAgentApp(name);
```
mohitpali, line 106: why allow overriding ?

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
blacktooth: We override this value in the serializer and de-serializer to replace any values explicitly set by configuration. Nevertheless, clients can still override and send garbage values if they really want to since we are just using the AWS SDK feature.

mohitpali (commented):
How have these been tested ?

common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaRegistryClient.java
blacktooth: Kept it short to not consume too many bits. No other specific reason. Lower case to be consistent with other components of the `UserAgent` 
```
vendor/Homebrew io/sync http/UrlConnection cfg/retry-mode/legacy
```

serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/GlueSchemaRegistryDeserializationFacade.java
blacktooth: This is required for KCL to set this explicitly. KCL doesn't have access to `GlueSchemaRegistryConfiguration`, so can't set it via configuration. [Ref](https://github.com/awslabs/amazon-kinesis-client/blob/master/amazon-kinesis-client/src/main/java/software/amazon/kinesis/schemaregistry/SchemaRegistryDecoder.java)

blacktooth:
> How have these been tested ?

This was tested using unit-tests and integration-tests. I have also manually tested this using example code. Verified that the logs contain the expected UserAgent strings.

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/UserAgents.java
mohitpali: We have anyways added a dependency there. A class unless loaded dynamically wont make a difference. Also, it gives us a streamlined option.

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
mohitpali: wouldn't that mess up our data. Can we not just protect it or allow only defined ENUMs. We can handle Kinesis case also by allowing KCL/KPL strings but checking if they are allowed ENUMs.

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
mohitpali: String.valueOf would also be consistent with rest of the class, but i guess this works too.

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
blacktooth: This library can also be used by different services like Lambda, Firehose etc. to serialize / deserialize. This config shouldn't become a bottleneck for adding a new enum. Instead the metrics can be derived from free-form text anyways.

PR #121, opened 2021-12-30
Ensure cached serializer schemas are not evicted in case of API failures.

blacktooth:
*Description of changes:*

Improving the caching mechanism to improve availability of the serializer. 

I will submit follow up for PR for de-serializer.

**Testing**

Unit tests and integration tests pass.


common/src/main/java/com/amazonaws/services/schemaregistry/common/SchemaByDefinitionFetcher.java
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @param schemaDefinition Schema Definition
+     * @param schemaName       Schema Name
+     * @param dataFormat       Data Format
+     * @param metadata         metadata for schema version
+     * @return Schema Version ID
+     * @throws AWSSchemaRegistryException on any error while fetching the schema version ID
+     */
+    @SneakyThrows
+    public UUID getORRegisterSchemaVersionId(
```
hhkkxxx133, line 65: Whats the benefit of changing to this?

blacktooth: If you meant, refactoring the class, it's for separating out the auto-registration logic so that the results can be cached. It wasn't possible earlier as this logic was part of AWSSchemaRegistryClient which is common for serializer and de-serializer. Adding a cache in AWSSchemaRegistryClient will instantiate it for both the serializer and de-serializer which is not optimal.

hhkkxxx133 (approved):

Issue #125, opened 2022-01-07
Strange schema issue with Kafka Connect


averemee-si:
In my Kafka Connect connector I have 2 different schemas for key and value data with different names:
```
			// Schema init - keySchema is immutable and always 1
			final SchemaBuilder keySchemaBuilder = SchemaBuilder
						.struct()
						.required()
						.name(tableFqn + ".Key")
						.version(1);
			final SchemaBuilder valueSchemaBuilder = SchemaBuilder
						.struct()
						.optional()
						.name(tableFqn + ".Value")
						.version(version);

```
SourceRecord is created using both schemas:
```
				keySchema = keySchemaBuilder.build();
				valueSchema = valueSchemaBuilder.build();
				//
				// Populations of Struct for key and value...
				//
				sourceRecord = new SourceRecord(
					sourcePartition,
					offset,
					kafkaTopic,
					keySchema,
					keyStruct,
					valueSchema,
					valueStruct);
```
My connect standalone.properties contains:
```
bootstrap.servers=localhost:9092

internal.key.converter=org.apache.kafka.connect.json.JsonConverter
internal.key.converter.schemas.enable=false
internal.value.converter=org.apache.kafka.connect.json.JsonConverter
internal.value.converter.schemas.enable=false

offset.storage.file.filename=/tmp/connect.offsets

offset.flush.interval.ms=10000
poll.interval.ms=5000

# Glue Schema Registry Specific Converters
key.converter=com.amazonaws.services.schemaregistry.kafkaconnect.AWSKafkaAvroConverter
key.converter.schemas.enable=true
value.converter=com.amazonaws.services.schemaregistry.kafkaconnect.AWSKafkaAvroConverter
value.converter.schemas.enable=true

# Converter-specific settings can be passed in by prefixing the Converter's setting
# with the converter we want to apply it to
key.converter.region=eu-north-1
key.converter.schemaAutoRegistrationEnabled=true
key.converter.avroRecordType=GENERIC_RECORD
# Default value: default-registry
key.converter.registry.name=TestSchemaRegistry001
value.converter.region=eu-north-1
value.converter.schemaAutoRegistrationEnabled=true
value.converter.avroRecordType=GENERIC_RECORD
# Default value: default-registry
value.converter.registry.name=TestSchemaRegistry001

```
This produces very strange schemas, two versions with same name in Glue Schema Registry (Kafka Connect schema names are different). Under version **1** - key schema:
```
{
  "type": "record",
[...]

PR #126, opened 2022-01-09
Add schema naming strategy for Kafka Connect

averemee-si:
*Issue #, if available:* 125
https://github.com/awslabs/aws-glue-schema-registry/issues/125

*Description of changes:*
Added a new implementation of AWSSchemaNamingStrategy that makes AWS Glue Schema names same as Kafka Connect schema names. Default AWSSchemaNamingStrategy implementation makes schema names same as topic name and this cause strange naming and versioning in AWS Glue Schema Registry for Kafka Connect records created using both key and value schemas.

### DIFF
--- a/common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaNamingStrategyConnectNameImpl.java
+++ b/common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaNamingStrategyConnectNameImpl.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.amazonaws.services.schemaregistry.common;
+
+import lombok.extern.slf4j.Slf4j;
+
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.Schema;
+
+@Slf4j
+public class AWSSchemaNamingStrategyConnectNameImpl implements AWSSchemaNamingStrategy {
+
+    /**
+     * Returns the schemaName.
+     * If data passed are in org.apache.avro.generic.GenericData$Record format
+     * data.getSchema().getFullName() is returned ('connect.name' of schema)
+     *
+     * @param transportName topic Name or stream name etc.
+     * @param data
+     * @return schema name.
+     */
+    @Override
+    public String getSchemaName(String transportName, Object data) {
+	if (data instanceof GenericData.Record) {
+	    return getConnectName((GenericData.Record) data, transportName);
+        } else {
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaNamingStrategyConnectNameImpl.java
```diff
@@ -0,0 +1,81 @@
...
+
+package com.amazonaws.services.schemaregistry.common;
+
+import lombok.extern.slf4j.Slf4j;
+
+import org.apache.avro.generic.GenericData;
+import org.apache.avro.Schema;
+
+@Slf4j
+public class AWSSchemaNamingStrategyConnectNameImpl implements AWSSchemaNamingStrategy {
```
blacktooth, line 24: Thanks for submitting this PR! 

Few comments:

1. In my opinion this is also useful for use-cases beyond Kafka connect. We can drop 'Connect' from the class name. Also, we support multiple data-formats like JSON / Protobuf (future) but logic here seems to be specific to Avro, I suggest following:
1.1 Add a TODO / issue and a note that this needs to be improved to support other data formats. I can take up the improvement later.
1.2 Feel free to implement it for JSON format if you can. :)

Name suggestion: `RecordSchemaNameNamingStrategy` or better.

2. Can we also support Avro SpecificRecord?  
3. In order to have predictable behavior, can we throw an exception if `fullName` is null or it's not an Avro record.
4. Were you able to test this end-to-end from serializer to de-serializer? It would be great if you can add a unit-test. 

Thanks again for the PR!

averemee-si: Thanks for review and superb suggestions!

1. I will rename it to AWSSchemaNamingStrategyRecordSchemaNameImpl
Unfortunately it seems to me that JSON Schema converter ignores value of both **key.converter.schemaNameGenerationClass** and **value.converter.schemaNameGenerationClass** parameters. So I'm unable to overload default implementation, unfortunately.
2. Yes, this works with **key.converter.avroRecordType=SPECIFIC_RECORD** and **value.converter.avroRecordType=SPECIFIC_RECORD**
3. Will be done.
4. Currently I'm testing in my Oracle RDBMS source and sink connector environment. I will try to backport some record examples for use in end-to-end unit test.

Regards,
Aleksei

blacktooth: 1. Thanks! We can drop `Impl` from the name. 
2. Should the code be updated as well? From the code, it seems like it only works if the passed object is `GenericRecord`.


Thank you!

MolochkoVitaly:
Hi @blacktooth @averemee-si
Could you please share status or any details on this PR?
It would be super cool to have this PR-feature in master

averemee-si:
> Hi @blacktooth @averemee-si
> Could you please share status or any details on this PR?
> It would be super cool to have this PR-feature in master

Hi Vitaly,

This naming strategy is also available in separate repo - [aws-glue-schema-naming](https://github.com/averemee-si/aws-glue-schema-naming)

Regards,
Aleksei

blacktooth:
@averemee-si Do you want to update this PR with suggested feedback?

common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaNamingStrategyConnectNameImpl.java
mohitpali: Thanks for this PR @averemee-si !

Would it be possible for you to implement the comments and resubmit this PR ?

Issue #120, opened 2021-12-28
Http client conflicts


samu-developments:
Hi,

I'm getting the error below when integrating schema registry in our project. 

```
Caused by: software.amazon.awssdk.core.exception.SdkClientException: Multiple HTTP implementations 
were found on the classpath. To avoid non-deterministic loading implementations, please explicitly provide 
an HTTP client via the client builders, set the software.amazon.awssdk.http.service.impl system property 
with the FQCN of the HTTP service to use as the default, or remove all but one HTTP implementation 
from the classpath
```

I cannot resolve this with the directions above (with our current httpClient), and believe the issue is that `UrlConnectionHttpClient` is used explicitly in the builder below. 

```
GlueClientBuilder glueClientBuilder = GlueClient
                .builder()
                .credentialsProvider(credentialsProvider)
                .overrideConfiguration(overrideConfiguration)
                .httpClient(UrlConnectionHttpClient.builder().build())
                .region(Region.of(glueSchemaRegistryConfiguration.getRegion()));

```
[source](https://github.com/awslabs/aws-glue-schema-registry/blob/35ad70de47e8af42c590e6a6f39b6b349e467e6b/common/src/main/java/com/amazonaws/services/schemaregistry/common/AWSSchemaRegistryClient.java#L93)

callado4:
It seems like it was added with this PR: https://github.com/awslabs/aws-glue-schema-registry/pull/49

@mohitpali is the UrlConnectionHttpClient needed for the JSON support? Have this dependency in the library conflicts with being able to use the ProfileCredentialProvider in the default configuration when this dependency is in the classpath.

blacktooth:
Thanks for reporting! I will investigate this and post my findings.

blacktooth:
I think we were specifying it explicitly to avoid conflicts with HttpClient. Are you constructing a AWS Client in your application? 
Did you try explicitly specifying the http client? 

Related [AWS SDK issue](https://github.com/aws/aws-sdk-java-v2/issues/446)

samu-developments:
You're right, found where it was implicitly created, creating it explicitly seems to have fixed the problem!

samu-developments:
I reopened the issue; I feel like the fact that it is explicitly specified here in this dependency should not affect the rest of the codebase having glue-schema-registry as a dependency. Instead of having to specify it explicitly everywhere else, glue-schema-registry should also use the service loader approach and support several http clients.

[...]

PR #127, opened 2022-01-11
remove url-http-client dependency

samu-developments:
*Issue #120*
*Description of changes:* 
Removes explicit use of UrlHttpClient, which causes conflicts when using other http clients in the same project. 


samu-developments:
Not necessary, see issue.

Issue #43, opened 2021-05-09
.NET support


einarjohnson:
Is anyone here familiar with the roadmaps for https://docs.aws.amazon.com/glue/latest/dg/schema-registry.html ?

It seems to me that Java is the only supported language at the moment? Are there any plans to introduce support for .NET and Python and other data formats such as Protobuf and JSON schema? Until this support is official, is there any recommended way to use .NET with the schema registry (direct API calls etc)

mohitpali:
Thanks for reaching out.

We do have plans to support multiple formats and languages. Some of them will be added this year. JsonSchema Support should get added by Q2. We are planning to add Protobuf support as well as Python and other languages support.

MichaelMt66:
Hi, I want to integrate my nodejs ms with Glue Schema Registry, any plans to introduce support to nodejs? and also, is there any recommended way to use nodejs?

mohitpali:
For node.js, you could use the aws sdk ( https://aws.amazon.com/sdk-for-javascript/ , https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-glue/classes/createschemacommand.html).

einarjohnson:
@blacktooth , is there any news on Protobuf support? Would be great to get some idea of when we can expect this to happen.

blacktooth:
@einarjohnson Protobuf support will GA in Q1 2022, please follow the status [here](https://github.com/awslabs/aws-glue-schema-registry/issues/34).  Pease feel free to get in touch with your AWS account team to grant you beta access.

yuoscer:
Hi @blacktooth, when will glue registry support golang?

mrfsbDev:
What is the plan to support other programming languages? We are intending to adopt the AWS Glue Schema Registry, our teams use different programming languages like Python, Go, NodeJS, but I am having a hard time understanding the SDKs, the examples are not objective, basically I am trying to serialize and deserialize to validating the schema, the SDKS has a lot of functionality, but the essentials of serializing and deserializing are not present or very confusing.

hhkkxxx133:
The multi-lang support is under active development currently. We will update once we have an ETA. Thanks!

morlando016:
> The multi-lang support is under active development currently. We will update once we have an ETA. Thanks!

Hi is there a roadmap or estimated quarter for this? My team is looking to use a .NET sdk.

suchitgupta01:
[...]

PR #167, opened 2022-04-27
Create module for native-schema-registry

blacktooth:
*Issue #, if available:*

#43 

*Description of changes:*

Creating module that builds a GraalVM native image of the existing Java SerDes.

**Note: This will be merged into a separate branch "native-schema-registry" instead of master**

### DIFF
--- a/native-schema-registry/pom.xml
+++ b/native-schema-registry/pom.xml
@@ -1,0 +1,154 @@
+<!--
+/*
+ * Copyright 2022 Amazon.com, Inc. or its affiliates.
+ * Licensed under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+  -->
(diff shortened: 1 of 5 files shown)


native-schema-registry/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+        <license>
+            <name>Apache License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.source>11</maven.compiler.source>
```
hhkkxxx133, line 53: Are we using this anywhere? The parent pom uses 1.8 for the source and target. Should we keep it consistent?

blacktooth: This module will use a different Java compiler than the parent. Since this will be compiled into a shared library, the version of JDK is not significant. I wanted to use the best stable JDK version supported by GraalVM.

hhkkxxx133: Will this introduce incompatibility issue for customers using JDK8?

YangSan0622: It should not be any issue, this is the java version used by GraalVM and it does not affect customer. In addition, customer are not using Java at all in this case since this shared lib is built for other languages such as Python/C# etc.

blacktooth: Also, the parent module shouldn't compile this sub-module as OpenJDK / Amazon Corretto won't be able to compile this. It will use API specific to GraalVM.

hhkkxxx133 (approved):
Please make sure this won't be merged into master.

blacktooth:
This is merged into a separate branch.

PR #191, opened 2022-06-17
Upgrading Avro version to prevent a CVE

YangSan0622:
*Description of changes:* Upgrading Avro version to 1.11.0 to prevent a CVE


blacktooth:
Please run integration tests and confirm this works fine.

YangSan0622:
I have ran the integration tests and they are all passing. We should be good.

PR #196, opened 2022-07-14
Create C library wrapper over existing Java library using GraalVM

blacktooth:
*Issue #, if available:*
#43 

*Description of changes:*

Port the Glue Schema Registry Java library into multiple languages (C#, Python, NodeJS, Go, etc.). The support is added by compiling the existing Java library using GraalVM into a C shared library (.so, .dll) and building Swig bindings on top of it.

*Design Choices*

1. **Memory de/allocation** - We use AWS SDK Common library for all explicit memory de/allocations. Memory of GraalVM / Java and target lang. are managed by those runtimes. 
2. **Exceptions** - In case of exceptions, `NULL` is returned and exception code / message are returned as a pointer holder to `glue_schema_registry_error` object. `mutable_byte_array * encode(..., glue_schema_registry_error **p_err)`. Swig intercepts this error object and throws as an exception in the target language.
3. **Platforms** - Java, C, target lang libraries are supported across Linux, OSX and Windows OSes on x86-64. ARM testing is pending.
4. **Testing** - We make use of `cmocka` to test the C code. There is no good ways to test GraalVM code as we cannot mock network calls. Target lang tests will test this code as well. We will add integration tests that cover all the use-cases. 
5. **Build** - CMake builds work across all the platforms. We use Google Address and Memory sanitizers to detect memory leaks and problems during builds.
6. **Language Bindings** - Swig generates all the language bindings code and CMAKE is used to compile them. The generated binaries are placed in corresponding language projects for those language build to use.

*Open questions*
1. Logging - The C code logs to `stderr`. We have to figure out how to pipe to target language logs.
2. Review usage of `aws-c-common` library.

*Testing*

Tested using unit-tests. Target language tests currently make network calls and need credentials to run. We will update our tests to use local docker based Glue service.

*Coverage*

Setup code coverage using `gcov` and build will fail if code coverage is less than 100.

*Linting*

clang-format and clang-tidy is setup to fail build if code formatting doesn't meet standards.

**This is one of the initial PRs to cover the data types and header files. Implementation PRs will follow.**


native-schema-registry/c/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+set(NATIVE_SCHEMA_REGISTRY_MODULE_NAME libnativeschemaregistry)
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+
+include_directories("include")
+include_directories(${LIB_NATIVE_SCHEMA_REGISTRY_PATH})
+
+add_subdirectory("src")
+include (CTest)
+enable_testing()
+add_subdirectory("test")
```
JonathanHenson, line 35: are you ever installing this as a standalone library? e.g. do you need to declare targets and config files?

native-schema-registry/c/include/glue_schema_registry_deserializer.h
```diff
@@ -0,0 +1,32 @@
+#ifndef GLUE_SCHEMA_REGISTRY_DESERIALIZER_H
+#define GLUE_SCHEMA_REGISTRY_DESERIALIZER_H
+
+#include "glue_schema_registry_schema.h"
+#include "glue_schema_registry_error.h"
+#include "mutable_byte_array.h"
+#include "read_only_byte_array.h"
+#include <stdbool.h>
+
+typedef struct glue_schema_registry_deserializer {
```
JonathanHenson, line 10: not blocking anything here, because it's 100% your call, but the typedef struct idiom goes off the rails when used too frequently. I tend to just use 'struct type' declarations.

native-schema-registry/c/include/glue_schema_registry_deserializer.h
```diff
@@ -0,0 +1,32 @@
...
+    //This is used for storing the instance context. Currently, being used for managing GraalVM instance.
+    void *instance_context;
+} glue_schema_registry_deserializer;
+
+glue_schema_registry_deserializer *new_glue_schema_registry_deserializer(glue_schema_registry_error **p_err);
+
+void delete_glue_schema_registry_deserializer(glue_schema_registry_deserializer *deserializer);
+
+mutable_byte_array *glue_schema_registry_deserializer_decode(glue_schema_registry_deserializer *deserializer,
+                                                             read_only_byte_array *array,
```
JonathanHenson, line 20: const?

native-schema-registry/c/include/glue_schema_registry_error.h
```diff
@@ -0,0 +1,54 @@
+#ifndef NATIVE_SCHEMA_REGISTRY_GLUE_SCHEMA_REGISTRY_ERROR_H
+#define NATIVE_SCHEMA_REGISTRY_GLUE_SCHEMA_REGISTRY_ERROR_H
+
+#include <stdio.h>
+
+//Error codes are arbitrarily listed from 5000. No specific reason.
+#define ERR_CODE_INVALID_STATE 5000
```
JonathanHenson, line 7: might want to namespace these.

native-schema-registry/c/include/glue_schema_registry_error.h
```diff
@@ -0,0 +1,54 @@
...
+#define ERR_CODE_NULL_PARAMETERS 5001
+#define ERR_CODE_GRAALVM_INIT_EXCEPTION 5002
+#define ERR_CODE_GRAALVM_TEARDOWN_EXCEPTION 5003
+#define ERR_CODE_INVALID_PARAMETERS 5004
+#define ERR_CODE_RUNTIME_ERROR 5005
+
+//TODO: Improve error reporting to respect logging levels.
+#define log_warn(msg, code) fprintf(stderr, "WARN: %s, Code: %d\n", msg, code)
```
JonathanHenson, line 15: namespace?

JonathanHenson, line 15: there's a whole logger in aws-c-common

native-schema-registry/c/include/glue_schema_registry_schema.h
```diff
@@ -0,0 +1,41 @@
...
+#define GLUE_SCHEMA_REGISTRY_SCHEMA_H
+
+#include "glue_schema_registry_error.h"
+
+/*
+ * Glue Schema Registry Schema structure that represents
+ * schema object required by Glue Schema Registry Serializers / De-serializers.
+ */
+typedef struct glue_schema_registry_schema {
+    //String name of the schema
```
JonathanHenson, line 11: don't use C++ comments in c header files if you can avoid it. Header files often need to be consumed by older language standards.

native-schema-registry/c/include/glue_schema_registry_schema.h
```diff
@@ -0,0 +1,41 @@
...
+
+//Gets different attributes from glue_schema_registry_schema instance.
+//These getter methods are translated into "Getter" methods in target languages.
+const char * glue_schema_registry_schema_get_schema_name(glue_schema_registry_schema * schema);
+
+const char * glue_schema_registry_schema_get_schema_def(glue_schema_registry_schema * schema);
+
+const char * glue_schema_registry_schema_get_data_format(glue_schema_registry_schema * schema);
+
+#endif //GLUE_SCHEMA_REGISTRY_SCHEMA_H
```
JonathanHenson, line 41: newline, I recommend running clang-format as a CI job as well.

native-schema-registry/c/include/glue_schema_registry_schema.h
```diff
@@ -0,0 +1,41 @@
...
+    const char * data_format,
+    glue_schema_registry_error ** p_err
+);
+
+//Deletes the glue schema registry schema.
+void delete_glue_schema_registry_schema(glue_schema_registry_schema * schema);
+
+//Gets different attributes from glue_schema_registry_schema instance.
+//These getter methods are translated into "Getter" methods in target languages.
+const char * glue_schema_registry_schema_get_schema_name(glue_schema_registry_schema * schema);
```
JonathanHenson, line 35: pick a lane on pointer declaration style.

native-schema-registry/c/include/mutable_byte_array.h
```diff
@@ -0,0 +1,45 @@
...
+#include <stdlib.h>
+#include "glue_schema_registry_error.h"
+
+//Integer.MAX_VALUE in Java
+//This gives ~2.1Gb limit on a record.
+#define MAX_BYTES_LIMIT 2147483647
+/**
+* A mutable byte array that allows write / updating bytes in a fixed array of size `max_len`.
+*/
+typedef struct mutable_byte_array {
```
JonathanHenson, line 12: if you're already using aws-c-common, why not use aws_byte_cursor and aws_byte_buf?

cursor is read only slice, and byte_buf is mutable.

native-schema-registry/c/include/read_only_byte_array.h
```diff
@@ -0,0 +1,35 @@
+#ifndef READ_ONLY_BYTE_ARRAY_H
+#define READ_ONLY_BYTE_ARRAY_H
+#include <stdlib.h>
+#include "glue_schema_registry_error.h"
+
+typedef struct read_only_byte_array {
```
JonathanHenson, line 6: I don't see what in the type system is making this read only?

native-schema-registry/c/include/read_only_byte_array.h
```diff
@@ -0,0 +1,35 @@
...
+ * Gets the reference to data pointed by this array.
+ */
+unsigned char * read_only_byte_array_get_data(read_only_byte_array * array);
+
+/**
+ * Gets the len of the data being pointed by this array.
+ */
+size_t read_only_byte_array_get_len(read_only_byte_array * array);
+
+#endif //READ_ONLY_BYTE_ARRAY_H
```
JonathanHenson, line 35: newline, clang-format would catch this.

native-schema-registry/c/src/memory_allocator.c
```diff
@@ -0,0 +1,21 @@
+#include "memory_allocator.h"
+#include "aws/common/allocator.h"
+
+/*
+ * Uses the default memory allocator from AWS SDK Common library to allocate / free memory.
+ */
+void *aws_common_malloc(size_t size) {
+    struct aws_allocator *allocator = aws_default_allocator();
```
JonathanHenson, line 8: aws_default_allocator doesn't change. So you could store it at file scope to save on function calls if you'd like. Though, to be fair, you're allocating memory so a single fn call is unlkely to be noticeable.

native-schema-registry/c/CMakeLists.txt
blacktooth: We don't have the need to install this as a standalone library as of now.

native-schema-registry/c/include/glue_schema_registry_deserializer.h
blacktooth: I will check how it plays with Swig and take a call.

native-schema-registry/c/include/glue_schema_registry_schema.h
blacktooth: Okay, I will update them.

native-schema-registry/c/include/glue_schema_registry_deserializer.h
blacktooth: It plays well with Swig as Swig generates the class structures in target languages from typefdef structs, https://www.swig.org/Doc4.0/SWIGDocumentation.html#SWIG_nn32
Leaving it as is.

native-schema-registry/c/include/glue_schema_registry_deserializer.h
blacktooth: Updated them.

native-schema-registry/c/include/glue_schema_registry_error.h
blacktooth: okay.

native-schema-registry/c/include/glue_schema_registry_error.h
blacktooth: Thanks for pointing it out. I will evaluate and take the logging part separately.

native-schema-registry/c/include/mutable_byte_array.h
blacktooth: Thanks for the suggestion. I have modified the `mutable_byte_array` to wrap a `aws_byte_buf` so that I don't do buffer management myself. The outer structure is still useful to have as it fits our exception handling and binding generation model.

I wanted to do the same by wrapping `aws_byte_cursor` under `read_only_byte_array` but couldn't find a method that would return a pointer to it (I might be missing something). I could only find this [method](https://github.com/awslabs/aws-c-common/blob/main/include/aws/common/byte_buf.h#L585).

native-schema-registry/c/src/memory_allocator.c
blacktooth: Are you suggestion to create a `static` member? We don't have a global initializer yet, so couldn't find a way to initialize it once.

native-schema-registry/c/include/read_only_byte_array.h
blacktooth: My intention is to create a structure that doesn't manage the memory of underlying array and provides a way to read it.

native-schema-registry/c/include/mutable_byte_array.h
JonathanHenson: You can return cursor.ptr in that case

native-schema-registry/c/src/memory_allocator.c
JonathanHenson: You can do it lazily which that function does under the hood anyways, or use AWS_thread_call_once

JonathanHenson (approved):
See comments above. I don’t have any blockers but I do recommend deciding on relevant to you questions before merging over.

PR #204, opened 2022-07-29
Explicitly call Deflater/Inflater#end to free native resources (#203)

### LINKED ISSUE #203: Excessive native memory usage when using the built-in compression
The current de-/compression code does not call [`java.util.zip.Deflater#end`](https://cr.openjdk.java.net/~iris/se/11/latestSpec/api/java.base/java/util/zip/Deflater.html#end()) or [`java.util.zip.Inflater#end`](https://cr.openjdk.java.net/~iris/se/11/latestSpec/api/java.base/java/util/zip/Inflater.html#end()). This can lead to excessive native memory usage as it leaves the removal of the internally allocated native memory buffers at the mercy of the JVM garbage collection and the finalizer kicking in which might be too late, i.e. resulting in processes being OOM killed.  We have observed applications configured with < 1GB max-heap requiring an additional 2GB of native memory and more.

mkr:
Fixes #203.


mohitpali (approved):
LGTM

Issue #220, opened 2022-10-03
Dependency on `everit-json-schema` 1.12.22 exposes vuln cve-2020-15250


neil-michalares:
Most recent version exhibits this behavior.

vulnerability:
https://nvd.nist.gov/vuln/detail/cve-2020-15250

source:
https://mvnrepository.com/artifact/org.json/json/20201115

dependency:
https://mvnrepository.com/artifact/com.github.erosb/everit-json-schema/1.12.2

blacktooth:
Hello, 

Thank you for your message about the CVEs reported in "aws-glue-schema-registry". After a thorough review we have determined this version is not impacted by the following CVEs:

CVE-CVE-2020-15250
CVE-2020-8908

blacktooth:
We can still pursue the PR to upgrade to latest versions.

YangSan0622:
Hello,
We have made a version upgrade and addressed this dependency vulnerability issue. The latest version released will not be impacted by this vulnerability.
Thank you so much for bringing this to our attention.

PR #221, opened 2022-10-03
Upgrade Evrit Json Schema to prevent cve-2020-15250

neil-michalares:
*Issue #, if available:*
Issue [220](https://github.com/awslabs/aws-glue-schema-registry/issues/220)
*Description of changes:*

bumping `everit.json.schema` to 1.14.1 to prevent [CVE](https://nvd.nist.gov/vuln/detail/cve-2020-15250)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,3 +59,6 @@ GlueSchemaRegistryKafkaSerializer/GlueSchemaRegistryKafkaDeserializer.
 
 ## Release 1.1.13
 * Upgraded kotlin dependency versions to prevent a CVE
+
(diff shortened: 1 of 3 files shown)


blacktooth:
Do we also have to update the `org.json.json` version?

neil-michalares:
@blacktooth I don't see a direct reference to that, so I don't believe we need to update anything else - however I did update the license version to match.

neil-michalares:
org.json:json looks to be brought in via this `everit.json.schema` dependency, so we should be in a good spot in terms of that listed CVE with this version bump.

PR #230, opened 2022-11-08
Update squareio wire-schema

strokyl:
*Issue*

This library depends on wireschema 3.7.1 when confluent protobuf deserializer depends on 4.3.0. This make it impossible to use both in the same project.
Here is my proposal to update wireschema to make both libraries compatible.
I didn't update to 4.4.3 (latest) because there is also breaking change between 4.3.0 and 4.4.0

*Description of changes:*

- use okio FakeSystem instead of jimfs ProtobufSchemaLoader (see: https://github.com/square/wire/pull/1918)

### DIFF
--- a/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/utils/apicurio/FileDescriptorUtils.java
+++ b/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/utils/apicurio/FileDescriptorUtils.java
@@ -329,6 +329,7 @@ public class FileDescriptorUtils {
     /**
      * When schema loader links the schema, it also includes google.protobuf types in it.
      * We want to ignore all the other types except for the ones that are present in the current file.
+     *
      * @return true if a type is a parent type, false otherwise.
      */
     private static boolean isParentLevelType(ProtoType protoType, Optional<String> optionalPackageName) {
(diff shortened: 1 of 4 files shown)


franoisD:
image

serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/utils/apicurio/FileDescriptorUtils.java
```diff
@@ -90,48 +90,48 @@ public class FileDescriptorUtils {
...
-            CalendarPeriodProto.getDescriptor().getFile(),
-            LocalizedTextProto.getDescriptor().getFile(),
-            IntervalProto.getDescriptor().getFile(),
-            ExprProto.getDescriptor().getFile(),
-            QuaternionProto.getDescriptor().getFile(),
-            PostalAddressProto.getDescriptor().getFile(),
-            ProtobufSchemaMetadata.getDescriptor().getFile(),
-            Decimals.getDescriptor().getFile()
+        return new FileDescriptor[]{
+                ApiProto.getDescriptor().getFile(),
```
YangSan0622, line 101: Can you revert these formatting change since it increases the difficultly of code review?

strokyl: Do you have any formatting tools in place to help me?

strokyl: Otherwise adding `&w=1` in the review URL will only show diff without space. https://github.blog/2018-05-01-ignore-white-space-in-code-review/

strokyl: I have just remove the whitespace anyway

strokyl:
up

strokyl:
up

wavejumper:
Just chiming in and saying that the use of 3.7.1 of wire-schema is also impacting our use of this library. We also have a dependency on Confluent's schema registry in our project. 
We depend on features in later versions of Confluent's schema registry, but can't bump because of this conflict.

Re CVE: while I understand from #235 that this library isn't impacted by this vulnerability, it would still be a plus to bump this dependency. For our use case, we have customers who consume our built artefact (a JAR file) and run their own CVE scanning on it. We try to strive to have as close to zero CVEs in our product.
Bumping wire-schema would also mean removing many suppressions from our nvd xml file (another win).

YangSan0622 (changes requested):
Sorry for the late update, I have went through the code and LGTM.
Do you mind resolve the merge confilct and squash the commits? I can approve the PR after that.

strokyl:
> Sorry for the late update, I have went through the code and LGTM. Do you mind resolve the merge confilct and squash the commits? I can approve the PR after that.

Done

YangSan0622 (approved):

PR #248, opened 2023-02-27
Lombok as provided dependency

### LINKED ISSUE #247: lombok shouldn't be transitive dependency
Lombok is an annotation processor. It is only needed at compile time so that it can perform the bytecode generation. It is not needed at runtime hence it must not be a transitive dependency. When it is on a classpath, even though a project that doesn't directly use it directly the compiler will still pick it up and try to execute it. This is usually a problem when old versions of lombok break on latest JDK versions. Developers waste time on trying to fix/workaround a problem which can be avoided if lombok was not brought in as a transitive depenedency.

Actual: lombok is a `comile` scope dependency
Expected: lombok is a `provided` scope dependency 

Quote from Lombok [documentation](https://projectlombok.org/setup/maven):

> To set up lombok with any build tool, you have to specify that the lombok dependency is required to compile your source code, but does not need to be present when running/testing/jarring/otherwise deploying your code. Generally this is called a 'provided' dependency.

NersesAM:
*Fixes #247*

*Description of changes:*

* Move lombok scope from compile to provided
* Remove lombok-utils as it doesn't exist in newer versions anymore

### DIFF
--- a/avro-flink-serde/pom.xml
+++ b/avro-flink-serde/pom.xml
@@ -68,6 +68,12 @@
             <artifactId>schema-registry-serde</artifactId>
             <version>1.1.15</version>
         </dependency>
+        <dependency>
+          <groupId>org.projectlombok</groupId>
+          <artifactId>lombok</artifactId>
+          <version>1.18.26</version>
+          <scope>provided</scope>
+        </dependency>
 
         <dependency>
             <groupId>software.amazon.glue</groupId>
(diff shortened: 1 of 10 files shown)


NersesAM:
@blacktooth can you please kick off the builds again? I have somehow managed to miss few files in the first commit.

mohitpali (approved):
LGTM

PR #251, opened 2023-03-08
Fix build failure on windows runners.

NersesAM:
build-tools doesn't need to be a dependency for the plugin as plugin config is referencing the files directly using absolute path. This is causing problem when the underlying .m2 directory is ephemeral between actions steps and the installed jar is lost which causes the next step (clean install) to fail. Also removing `mvn test` step as per maven lifecycle previous `mvn install` already runs the `test` phase.

### DIFF
--- a/.github/workflows/MavenCI.yml
+++ b/.github/workflows/MavenCI.yml
@@ -6,9 +6,9 @@ on:
   pull_request:
     branches: [ master ]
   release:
-    branches: [ master ] 
-    types: [ created ]  
-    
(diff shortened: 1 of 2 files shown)


mohitpali (approved):

mohitpali:
Thanks for making this change. As a fix I am merging this. However, please also validate if checkstyle is succeeding.

NersesAM:
@mohitpali it is working. you can see this in the master build currently running

image

PR #246, opened 2023-02-26
Add property to provide a http proxy url to connect to glue registry

abhilash47:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #245, opened 2023-02-10
How to provide http proxy configuration to AwsSchemaRegistryClient ?


abhilash47:
I am trying to connect to aws glue registry from a corporate network. To be able to connect to glue registry I need to go through proxy. Without a proxy, the connection is timing out. 

I am able to build a GlueClient by passing http proxy and able to do get schema by calling getSchemaVersion(). I have passed proxy configuration for glue client as below 

@Bean
public GlueClient glueClient() {
	return GlueClient.builder()
		.region(Region.EU_WEST_1)
		.httpClient(apacheHttpClient())
		.build();
}

@Bean
public SdkHttpClient apacheHttpClient() {
         return ApacheHttpClient.builder()
		.proxyConfiguration(proxyConfiguration())
		.build();
}

@Bean
public ProxyConfiguration proxyConfiguration() {
         return ProxyConfiguration.builder()
		.endpoint(URI.create("http://proxy-url:proxy-port"))
		.username("proxy-user")
		.password("proxy-password")
		.build();
}

But when I try to publish a message, the kafka producer is timing out while waiting for AwsSchemaRegistryClient to get schema. 
Below is the error. 

023-02-10 14:39:21.486 DEBUG 17268 --- [nio-8888-exec-2] software.amazon.awssdk.request           : Retryable error detected. Will retry in 177ms. Request attempt number 2

software.amazon.awssdk.core.exception.SdkClientException: Unable to execute HTTP request: connect timed out
	at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:102) ~[sdk-core-2.17.148.jar:na]
	at software.amazon.awssdk.core.exception.SdkClientException.create(SdkClientException.java:47) ~[sdk-core-2.17.148.jar:na]
	at software.amazon.awssdk.core.internal.http.pipeline.stages.utils.RetryableStageHelper.setLastException(RetryableStageHelper.java:204) ~[sdk-core-2.17.148.jar:na]
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:83) ~[sdk-core-2.17.148.jar:na]
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:36) ~[sdk-core-2.17.148.jar:na]
[...]

PR #49, opened 2021-06-28
Add JSON Support

mohitpali:
*Issue #, if available:*

*Description of changes:* Add support for JSON Data Format with JSON Schema defined by https://json-schema.org/

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #263, opened 2023-05-17
Remove url-connector http lib dependency, require GlueClientBuilder input to AWSSchemaRegistryClient

dbottini:
Issue: https://github.com/awslabs/aws-glue-schema-registry/issues/120 (Primary)

Description: 
AWSSchemaRegistryClient in the `common` module directly imports the UrlConnectionHttpClient, which can cause classpath friction with AWS SDK in other places where ApacheHttpClient is preferred, and the ServiceLoader can't figure out which client to use. An example of this is when using the Kinesis Client Library v2, their addition of this library pollutes the classpath, when it ideally should've waited for a user to deliberately configure in the UrlConnectionHttpClient through a GlueClient. This change prefers rely on the GlueClient defaults, or gives the user _full_ access to the GlueClient builder. It still reserves the right to add some overrides around endpoint, region, etc.

This does break the bespoke integration with the UrlConnection Proxy Url, and now requires manually configuring the Glue Client. This may further impact scenarios where Proxy Url was specified using one of the higher-order libraries around `common`, where the user expects GlueSchemaRegistryConfiguration to set everything up. Now, to create an object like `GlueSchemaRegistrySerializationFacade`, the user has to also create `SchemaByDefinitionFetcher` which then wraps `AWSSchemaRegistryClient` which then wraps `GlueClient` which contains a manually configured UrlConnectionHttpClient with a proxy url.

The url-connection http lib was likely unintentionally added in https://github.com/awslabs/aws-glue-schema-registry/pull/49/files#diff-79d226ed87c480fcfac3a55dfae33d93c29b6b43f7eebbb1369a4557581431bd, and the proxy url functionality was cemented on in https://github.com/awslabs/aws-glue-schema-registry/pull/246. It's worth noting that the initial issue, https://github.com/awslabs/aws-glue-schema-registry/issues/245, was attempting to directly insert a GlueClient, so this is more cohesive with their initial approach.

Related reports of issue:
https://github.com/awslabs/aws-glue-schema-registry/pull/127
https://stackoverflow.com/a/75323773

### DIFF
--- a/common/src/test/java/com/amazonaws/services/schemaregistry/common/AWSSchemaRegistryClientTest.java
+++ b/common/src/test/java/com/amazonaws/services/schemaregistry/common/AWSSchemaRegistryClientTest.java
@@ -140,13 +140,6 @@ public class AWSSchemaRegistryClientTest {
     }
 
     @Test
-    public void testConstructor_nullCredentials_throwsException() {
-        glueSchemaRegistryConfiguration = new GlueSchemaRegistryConfiguration(configs);
-        Assertions.assertThrows(IllegalArgumentException.class , () -> new AWSSchemaRegistryClient(null,
-            glueSchemaRegistryConfiguration));
-    }
-
-    @Test
     public void testConstructor_nullSerdeConfigs_throwsException() {
         AwsCredentialsProvider mockAwsCredentialsProvider = mock(AwsCredentialsProvider.class);
         Assertions.assertThrows(IllegalArgumentException.class , () -> new AWSSchemaRegistryClient(mockAwsCredentialsProvider,null ));
@@ -154,7 +147,7 @@ public class AWSSchemaRegistryClientTest {
 
     @Test
     public void testConstructor_nullGlueClient_throwsException() {
-        Assertions.assertThrows(IllegalArgumentException.class , () -> new AWSSchemaRegistryClient(null));
+        Assertions.assertThrows(IllegalArgumentException.class , () -> new AWSSchemaRegistryClient((GlueClient) null));
     }
 
     @Test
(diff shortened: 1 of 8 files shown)


common/src/test/java/com/amazonaws/services/schemaregistry/common/AWSSchemaRegistryClientTest.java
```diff
@@ -139,13 +139,6 @@ public void testAWSSchemaRegistryClient_putSchemaVersionMetadata_succeeds() {
         }
     }
 
-    @Test
```
dbottini, line 142: This test should be unnecessary - we should trust the default credentials provider.

dbottini:
It seems there's some old code from my initial attempt before a refactor. I also knocked out one of the constructors - I'll tidy those up.

serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/GlueSchemaRegistrySerializationFacade.java
```diff
@@ -65,8 +64,7 @@ public GlueSchemaRegistrySerializationFacade(@NonNull AwsCredentialsProvider cre
 
         if (this.schemaByDefinitionFetcher == null) {
             final AWSSchemaRegistryClient client =
-                new AWSSchemaRegistryClient(credentialProvider, this.glueSchemaRegistryConfiguration,
-                    AWSSchemaRegistryGlueClientRetryPolicyHelper.getRetryPolicy());
```
dbottini, line 69: retry policy will get automatically filled in by the less-telescoped constructor

dbottini:
One thing I'm noting here working with this repo is that there's a _lot_ of redundant dependency declarations in the child poms, which leads to errors like leaving lombok as a runtime dependency when it wanted to be provided. I'm not going to boil the ocean in this PR, but did want to clean up lombok at least. I think in the long term, this repo should consider moving the "all children want this dependency" dependencies to a basic `<dependency>` block rather than `<dependencyManagement>` to reduce risk of similar errors in the future. Probably every compile-time library, like lombok, should just be globally provided and not included individually.

PR #276, opened 2023-07-18
Update dependencies to resolve vulnerabilities from transitive dependencies

YangSan0622:
Motivation: Update dependencies for security purpose

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/serializer-deserializer/pom.xml
+++ b/serializer-deserializer/pom.xml
@@ -104,6 +104,13 @@
         <dependency>
             <groupId>com.github.erosb</groupId>
             <artifactId>everit-json-schema</artifactId>
+            <version>1.14.2</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-collections</groupId>
+                    <artifactId>commons-collections</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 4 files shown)


serializer-deserializer/pom.xml
```diff
@@ -221,6 +232,12 @@
             <groupId>com.google.protobuf</groupId>
             <artifactId>protobuf-java</artifactId>
         </dependency>
+        <!-- Temporarily adding dependency on a transitive dependency to fix security bug in underlying library. Remove when dependency is upgraded. -->
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
```
blacktooth, line 238: Do we need to remove the old collections dependency? Or is it a transitive dependency?

blacktooth:
Does the dependency tree show that these vulnerable dependencies are removed?

YangSan0622:
Intellij has a dependency checker(powered by Chekmarx), I confirmed the these vulnerabilities disappeared there

blacktooth (approved):

serializer-deserializer/pom.xml
YangSan0622: it is a transitive dependency, I have excluded it

PR #280, opened 2023-07-26
Use default region provider chain

### LINKED ISSUE #279: Using the default region provider chain
The region property is required using `AWSSchemaRegistryConstants.AWS_REGION`. However, AWS SDKs (both v1 and v2) provide a default region provider chain. Can the SerDes classes not select the region based on the chain? `AWSSchemaRegistryConstants.AWS_REGION` can still serve as override instead of being required. This also maintains backwards compatibility.

singhbaljit:
Closes #279 

If the `AWSSchemaRegistryConstants.AWS_REGION` property is not present, it attempts to use `DefaultAwsRegionProviderChain` to determine the AWS region. This is backwards compatible: the `AWSSchemaRegistryConstants.AWS_REGION` takes precedence, and throws the `AWSSchemaRegistryException` if unable to determine the region.

### DIFF
--- a/common/src/test/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfigurationTest.java
+++ b/common/src/test/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfigurationTest.java
@@ -32,6 +32,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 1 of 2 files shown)


common/src/test/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfigurationTest.java
```diff
@@ -103,12 +104,28 @@ public void testBuildConfig_noRegionConfigsSupplied_throwsException() {
         assertEquals("Region is not defined in the properties", exception.getMessage());
     }
 
+    /**
```
mohitpali, line 107: Thanks for this PR. May I suggest you add a negative test case also where DefaultRegionProviderChain throws SDKClientException ?

mohitpali (changes requested):

common/src/test/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfigurationTest.java
singhbaljit: The negative is already there. See [`testBuildConfig_noRegionConfigsSupplied_throwsException`](https://github.com/awslabs/aws-glue-schema-registry/blob/master/common/src/test/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfigurationTest.java#L96).

singhbaljit: @mohitpali Gentle reminder to please have another look.

mohitpali (approved):

PR #286, opened 2023-08-03
Remove AWS SDK v1 dependencies from the serde module

junyuc25:
Removing AWS SDK v1 dependencies. 

*Description of changes:*
In 2024, the SDK v1 will transition into [maintenance mode](https://docs.aws.amazon.com/sdkref/latest/guide/maint-policy.html) where only critical updates will be issued. The process of full deprecation is slated to begin in early 2025. 

The SDK v2 brings [several enhancements](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/home.html) for both customers and applications. These include non-blocking I/O and the option to plug-in different HTTP implementations at runtime. Thus to reduce maintenance overhead, we should migrate to SDK v2 as soon as possible.

### DIFF
--- a/avro-flink-serde/pom.xml
+++ b/avro-flink-serde/pom.xml
@@ -21,7 +21,7 @@
 
     <groupId>software.amazon.glue</groupId>
     <artifactId>schema-registry-flink-serde</artifactId>
-    <version>1.1.16</version>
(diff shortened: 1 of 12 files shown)


junyuc25:
cc @blacktooth @mohitpali

Issue #69, opened 2021-08-12
Using AWS signature for authentication


dexter-mh-lee:
Hi. 

I use IAM service accounts in k8s to give correct permissions to each pod talking to AWS systems. I could not find a documentation for how to do this for Glue schema registry. Is there a plan to add this functionality?

Currently, I need to add the permission to the worker node's IAM role, which is not ideal.

liko9:
@dexter-mh-lee What sort of Kubernetes are you using?  What process are you using to apply the policies to your IAM service accounts?

dexter-mh-lee:
We are using AWS EKS, and usually for our other use cases (Elasticsearch, MySQL, and so on), we can create a IAM service account with the permissions to access the AWS resources, and the AWS Signature v4 APIs run on each pod automatically picks up these policies (from the service account attached to the pod) and decides whether the pod has permissions to do certain actions. We couldn't find anything like this for Glue schema registry, so we had to give such permissions to the worker nodes of our EKS cluster directly, which is not ideal.

liko9:
@dexter-mh-lee Did you see these?  https://docs.aws.amazon.com/glue/latest/dg/using-identity-based-policies.html#access-policy-examples-aws-managed (specifically the AWSGlueSchemaRegistryReadonlyAccess managed policy, which feels like a good fit for your use case)

dexter-mh-lee:
We are already using this. Problem is that I have to give such permission to our EKS worker nodes, meaning every single pod in the cluster has access to the schema registry. This is against the principles used by the rest of the AWS ecosystem.

liko9:
I apologize, I am not an expert in IAM - do you have a link to the process/documentation you are using for your other use cases where you are able to apply this at a pod level?  I'm struggling to understand why you cannot apply this only to specifically labeled pods rather than the entire node.

dexter-mh-lee:
https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

This is the usual way of attaching roles to pods.

liko9:
My humblest apologies for not responding sooner.  I think it might be best for you to open a support case so we can investigate what is happening on the service side of both EKS and Glue Schema Registry to see if there isn't a solution.

blacktooth:
As suggested by Liko, you can get better help on this by reaching out to EKS AWS support team. Please let us know if there is something specific to GSR is not working. Closing this for now, please feel free to re-open or open a new issue if required. 

[...]

PR #303, opened 2023-10-12
use Apache HTTP client

### LINKED ISSUE #120: Http client conflicts

singhbaljit:
*Issue #, if available:*

Fixes #120, and possibly #69.

*Description of changes:*

Uses the Apache HTTP client instead of the URL client. Apache is the default client for AWS SDK v2 clients. Otherwise, it forces users to define explicitly define HTTP clients in their own code, and also breaks STS authentication in the default credentials provider.


singhbaljit:
@blacktooth Any thoughts?

Issue #307, opened 2023-11-08
Kafka Connect converter fails when schema contains null default values for record field


jurgispods:
Suppose we have the following Avro schema:

```
{
   "name":"data",
   "type":{
      "type":"record",
      "name":"Product",
      "fields":[
         {
            "name":"name",
            "type":{
               "type":"record",
               "name":"LanguageString",
               "fields":[
                  {
                     "name":"de",
                     "type":{
                        "type":"string",
                        "avro.java.string":"String"
                     }
                  },
                  {
                     "name":"en",
                     "type":[
                        "null",
                        {
                           "type":"string",
                           "avro.java.string":"String"
                        }
                     ]
                  }
               ]
            },
            "default":{
               "de":"",
               "en":null
            }
         }
      ]
   }
}
```

Note the `default` value that is specified as an object. 

Using `AWSKafkaAvroConverter` inside a Kafka Connect connector fails:

```
org.apache.kafka.connect.errors.ConnectException: Tolerance exceeded in error handler
	at org.apache.kafka.connect.runtime.errors.RetryWithToleranceOperator.execAndHandleError(RetryWithToleranceOperator.java:206)
	at org.apache.kafka.connect.runtime.errors.RetryWithToleranceOperator.execute(RetryWithToleranceOperator.java:132)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.convertAndTransformRecord(WorkerSinkTask.java:516)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.convertMessages(WorkerSinkTask.java:493)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.poll(WorkerSinkTask.java:332)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.iteration(WorkerSinkTask.java:234)
	at org.apache.kafka.connect.runtime.WorkerSinkTask.execute(WorkerSinkTask.java:203)
	at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:188)
	at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:243)
	at java.base\/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base\/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base\/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[...]

PR #308, opened 2023-11-08
fix issue with Avro record default values in Kafka Connect converter

jurgispods:
*Issue #, if available:* #307

*Description of changes:* See issue for description of the problem. This PR adds a check for `JsonProperties.NULL_VALUE` in addition to `null` to fix this problem.

### DIFF
--- a/avro-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/avrodata/AvroData.java
+++ b/avro-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/avrodata/AvroData.java
@@ -1225,7 +1225,7 @@ public class AvroData {
     private Object toConnectData(Schema schema, Object value, ToConnectContext toConnectContext,
                                  boolean doLogicalConversion) {
         validateSchemaValue(schema, value);
-        if (value == null) {
+        if (value == null || value == JsonProperties.NULL_VALUE) {
             return null;
         }
         try {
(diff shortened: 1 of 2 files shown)


jurgispods:
Did anyone have a chance to look at this?

YangSan0622:
Can you add unit test for this feature and ensure the existing integration test is passing after this change?

jurgispods:
@YangSan0622 I copied the same unit test used in https://github.com/confluentinc/schema-registry/pull/1928 - `mvn test` runs successfully.

YangSan0622:
Have you run the integration tests in https://github.com/awslabs/aws-glue-schema-registry/tree/master/integration-tests?

jurgispods:
@YangSan0622  Running `mvn test` from the `integration-test` directory runs through, but `run-local-tests.sh` fails with a few "Access denied" errors. Is there any documentation on how to run the tests?  It seems the tests need an AWS account with certain permissions. If that is the case, it would be very helpful if you could run the tests in your environment.

jurgispods:
Hi, any update on this? I will gladly run the integration tests, given I know how to do so and I have permission to access the corresponding AWS resources. Otherwise, I am dependent on the help of a maintainer.

jvdadda:
Hi @jurgispods, you had this right back in 2023: Avro represents a JSON `null` default as the `JsonProperties.NULL_VALUE` singleton, and `AvroData` only checked for a Java `null`. The same fix, with the check also applied in `validateSchemaValue` so that a non-optional schema still rejects it, is in our maintained fork of the Java library as https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/68, with your #307 schema as the regression test. It is available in the latest published release, currently 4.0.0.

It is on Maven Central as `com.mobsuccess:schema-registry-kafkaconnect-converter`, same packages and configuration keys; Connect workers take the `-plugin.zip` attached to each release instead. On the question you never got an answer to: the fork's unit suite (2000+ tests, inherited from here and kept complete) runs on every PR on JDK 17, 21 and 25, so a contributor does not need an AWS account to get a change validated.

PR #315, opened 2023-12-19
Append 2.12 suffix to artifactId for all modules

junyuc25:
*Issue #, if available:*

*Description of changes:*
This change allows glue-schema-registry to publish different version of artifacts based on scala version. With profile scala-2.12, it would compile with mbknor-jackson-jsonschema_2.12 and kafka_2.12; With profile scala-2.13, it would compile with mbknor-jackson-jsonschema_2.13 and kafka_2.13.

  * Append 2.12 suffix to artifactId for all modules
  * Add two maven profiles for Scala 2.12 and 2.13
  * Add a script to change the scala version for all modules

### DIFF
--- a/avro-flink-serde/pom.xml
+++ b/avro-flink-serde/pom.xml
@@ -20,7 +20,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>software.amazon.glue</groupId>
-    <artifactId>schema-registry-flink-serde</artifactId>
+    <artifactId>schema-registry-flink-serde_2.12</artifactId>
     <version>1.1.17</version>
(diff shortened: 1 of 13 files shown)


jurgispods:
I'm very interested in this PR, as it would allow us to use the aws-glue-schema-registry libs (using Scala 2.12 dependencies) inside the official Confluent Kafka Docker images (using Scala 2.13).

francesconero:
I'm very interested in this PR, as it would allow using this library in Scala 2.13 projects without dependency conflicts. Currently, the 2.12 hardcoded dependencies cause conflicts when our project already depends on the 2.13 versions of the same libraries.

Are there currently any blockers or is it just not high on the priority list?

Issue #281, opened 2023-07-26
Unable deserialize Protobuf POJO with Spring Boot


singhbaljit:
I've the following configuration:

```properties
spring.kafka.consumer.value-deserializer=com.amazonaws.services.schemaregistry.deserializers.GlueSchemaRegistryKafkaDeserializer
spring.kafka.properties.dataFormat=PROTOBUF
spring.kafka.properties.protobufMessageType=POJO
```

```java
  @KafkaListener(groupId = "test", topics = "test")
  void consume(final Timestamp payload)  {
  }
```

However, it looks like the descriptor based approach isn't detecting the class to to be `com.google.protobuf.Timestamp`; instead it thinks the class is `com.google.protobuf.Any`. I get the following error:
> org.springframework.messaging.converter.MessageConversionException: Cannot convert from [com.google.protobuf.Any] to [com.google.protobuf.Timestamp]

singhbaljit:
I do prefer this detection approach over having a configuration (i.e. (`KafkaProtobufDeserializerConfig.SPECIFIC_PROTOBUF_VALUE_TYPE` in Confluent SerDes).

singhbaljit:
Debugging further inside [ProtobufDeserializer#deserialize](https://github.com/awslabs/aws-glue-schema-registry/blob/master/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/protobuf/ProtobufDeserializer.java#L64), it looks like the `schema` and `schemaDefinition` are both correctly pointing to `com.google.protobuf.Timestamp`.

PR #336, opened 2024-02-29
fix(protobuf): Use the current thread classloader when deserializing …

### LINKED ISSUE #186: protobuf deserialize as pojo failure with quarkus
When you try to deserialize a protobuf message inside of kafka streams using quarkus, the quarkus class loader has layered loading which causes Class.forName to fail and be blocked.  It should be using the current thread class loader.

Currently ProtobufWireFormatDecoder.java at line 72 is
            final Class<?> classType = Class.forName(className);
But to make it compatible with layered class loaders it should be
            final Class<?> classType = Thread.currentThread().getContextClassLoader().loadClass(className);

danprueitt:
…to pojo.

*Issue #, if available:*

Closes #186

*Description of changes:*
Use the current thread classloader when deserializing protobuf to pojo to add support for layered classloaders, such as what quarkus uses in dev mode.

### DIFF
--- a/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/protobuf/ProtobufWireFormatDecoder.java
+++ b/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/protobuf/ProtobufWireFormatDecoder.java
@@ -69,7 +69,7 @@ public class ProtobufWireFormatDecoder {
     private Object deserializeToPojo(final Descriptors.Descriptor descriptor, final CodedInputStream codedInputStream) {
         final String className = ProtobufClassName.from(descriptor);
         try {
-            final Class<?> classType = Class.forName(className);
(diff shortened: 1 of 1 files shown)


jvdadda:
Hi @danprueitt, this one-liner from 2024 is the fix behind a whole family of reports (#186 Quarkus, #281 Spring Boot, #364 repackaged JAR, and #339), and it deserved better than sitting unreviewed. We ported it, credited to this PR, into our maintained fork of the Java library in https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/113. It is available in the latest published release, currently 4.0.0, on Maven Central as `com.mobsuccess:schema-registry-serde` (same packages and configuration keys).

Two small extensions over your patch, for the record: resolution asks the context class loader first and falls back to `Class.forName` when that loader is absent or cannot see the class, so it can only find more classes than before, never fewer; and the JSON `className` path in `JsonDeserializer` had the identical defect, so it goes through the same resolver.

If upstream ever picks back up, we're happy to contribute this back.

Issue #317, opened 2023-12-22
schema-registry-serde:1.1.17  transient dependency org.json:json:jar:20230227 CVE-2023-5072


fredrikls:
Currently:
software.amazon.glue:schema-registry-serde:jar:1.1.17 → com.github.erosb:everit-json-schema:jar:1.14.2 → org.json:json:jar:20230227

Fix:
Fixed in com.github.erosb:everit-json-schema:jar:1.14.3 -> org.json:json:jar:20231013

lapostoj:
It looks like this was released on the January one so it probably can be closed?
https://github.com/awslabs/aws-glue-schema-registry/releases/tag/v1.1.18

yhgillet:
We see the same issue also, for some reason the transitive version shown in intellij doesnt match the one we see in our project: 
<img width="355" alt="image" src="https://github.com/awslabs/aws-glue-schema-registry/assets/49912488/6186be06-1677-4315-967b-0576f85542e6">

yhgillet:
Any updates ? I've created a PR that should fix it

imvtsl:
Hey @blacktooth, @hiimanii, @YangSan0622 
The [PR](https://github.com/awslabs/aws-glue-schema-registry/pull/342) created by @yhgillet is a temporary fix. Can I go ahead and create a new PR for permanent fix.

The proposed PR will include the following update:
`com.github.erosb:everit-json-schema:jar` has new [release](https://github.com/everit-org/json-schema/releases/tag/1.14.3) to fix this vulnerability. We can bump dependency from 1.14.2 to its latest version, that is, [1.14.5](https://github.com/everit-org/json-schema/releases/tag/1.14.5).

This fix is necessary for me because I am using aws-glue-schema-registry, and vulnerability scan is flagging this CVE in this transitive dependency.

imvtsl:
Seems @bowen-s created a [PR](https://github.com/awslabs/aws-glue-schema-registry/pull/365) with this proposed change.

bowen-s:
The fix is merged in with the new release. Thanks for the issue. We will be releasing it on Maven shortly

imvtsl:
I noticed that release `1.1.23`, which includes this fix, is now available on Maven. This issue can now be closed.

PR #342, opened 2024-04-26
Fix for CVE-2023-5072 with transitive json dependency

yhgillet:
Issue #, if available:
#317 CVE-2023-5072

Description of changes:

Force dependency on latest json version instead of relying on transitive that has an older version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,6 +116,7 @@
         <protobuf.java.version>3.19.6</protobuf.java.version>
         <localstack.utils>0.2.23</localstack.utils>
         <aws.msk.iam.auth>2.0.3</aws.msk.iam.auth>
+        <org.json.version>20231013</org.json.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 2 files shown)


imvtsl:
@yhgillet 
Now that `com.github.erosb:everit-json-schema:jar` [update](https://github.com/everit-org/json-schema/releases/tag/1.14.3) is available, we can bump it from `1.14.2` to `1.14.3` instead of this temporary fix?

Please let me know if you want to make change in this PR. Otherwise, I would be happy to contribute.

imvtsl:
> @yhgillet Now that `com.github.erosb:everit-json-schema:jar` [update](https://github.com/everit-org/json-schema/releases/tag/1.14.3) is available, we can bump dependency from `1.14.2` to `1.14.3` instead of this temporary fix?
> 
> Please let me know if you want to make change in this PR. Otherwise, I would be happy to contribute.

`1.14.4` is the latest version available. We might as well update to this latest version.

bowen-s:
The fix is merged in with the new release [1.1.23](https://github.com/awslabs/aws-glue-schema-registry/releases/tag/v1.1.23). Thanks for the PR.

Issue #371, opened 2025-03-21
Dependency clashes on newer version Flink if "flink-avro:1.19.2" is in classpath


bing-hsu:
`software.amazon.glue:schema-registry-flink-serde:1.1.23` is compiled against `org.apache.flink:flink-avro:1.12.2`.

When using a Flink runtime of a newer version - `1.19.2` with `org.apache.flink:flink-avro:1.19.2` in the classpath, job fails reporting `java.lang.NoSuchMethodError`:

```txt
Caused by: java.lang.NoSuchMethodError: 'org.apache.avro.io.BinaryEncoder com.amazonaws.services.schemaregistry.flink.avro.GlueSchemaRegistryAvroSerializationSchema.getEncoder()'
	at com.amazonaws.services.schemaregistry.flink.avro.GlueSchemaRegistryAvroSerializationSchema.serialize(GlueSchemaRegistryAvroSerializationSchema.java:106)
	at org.apache.flink.connector.kafka.sink.KafkaRecordSerializationSchemaBuilder$KafkaRecordSerializationSchemaWrapper.serialize(KafkaRecordSerializationSchemaBuilder.java:348)
	at org.apache.flink.connector.kafka.sink.KafkaWriter.write(KafkaWriter.java:193)
```

If `org.apache.flink:flink-avro:1.19.2` is removed then job works as expected.

darenwkt:
Hi, I took a look at this issue, and as you have rightly pointed out, since flink 1.19 onwards, there is a change to getEncorder method return type in flink-formats/flink-avro/src/main/java/org/apache/flink/formats/avro/AvroSerializationSchema.java in https://github.com/apache/flink/commit/f8da7589223de8fd2a0809595ce33f217de6305f#diff-6da958ea13dc42cb06d10e96c86455ab15a02721d6f5f187cb0ab6f8cf0115dfL110

The return type has changed from BinaryEncoder to Encoder which is a breaking change.

I have attempted to fix this by bumping the flink version in https://github.com/awslabs/aws-glue-schema-registry/pull/374

jvdadda:
[...]

PR #374, opened 2025-05-06
Bump flink version to 1.20.1

darenwkt:
*Issue #, if available:*
https://github.com/awslabs/aws-glue-schema-registry/issues/371

Since flink 1.19 onwards, there is a change to `getEncorder` method return type in `flink-formats/flink-avro/src/main/java/org/apache/flink/formats/avro/AvroSerializationSchema.java` in https://github.com/apache/flink/commit/f8da7589223de8fd2a0809595ce33f217de6305f#diff-6da958ea13dc42cb06d10e96c86455ab15a02721d6f5f187cb0ab6f8cf0115dfL110

The return type has changed from `BinaryEncoder` to `Encoder`. 

As software.amazon.glue:schema-registry-flink-serde:1.1.23 is compiled against org.apache.flink:flink-avro:1.12.2, this means that flink 1.19 onwards no longer works with software.amazon.glue:schema-registry-flink-serde:1.1.23 and throws java.lang.NoSuchMethodError exception.

*Description of changes:*

This change bumps the flink dependency version from 1.12.2 to 1.20.1. 

Note that this is a breaking change as older flink version will not work with this new flink-avro version, hence recommend this commit to be included in a new major version release

### DIFF
--- a/avro-flink-serde/pom.xml
+++ b/avro-flink-serde/pom.xml
@@ -141,7 +141,7 @@
         <dependency>
             <groupId>org.apache.flink</groupId>
             <artifactId>flink-avro</artifactId>
-            <version>1.12.2</version>
+            <version>1.20.1</version>
             <exclusions>
                 <exclusion>
                     <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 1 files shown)


jvdadda:
Hi @darenwkt, your diagnosis on #371 was exactly right, and this bump is the fix. Our maintained fork of the Java library made the same move in https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/164, at 1.20.5 rather than 1.20.1 and on the suffix-less `flink-streaming-java` coordinate, since the `_2.11` line the pom pointed at ended with Flink 1.14. No source change was needed; the artifact simply needs a Flink 1.19 or later runtime, because of the `getEncoder()` return type you traced. It is available in the latest published release, currently 4.1.0, on Maven Central as `com.mobsuccess:schema-registry-flink-serde`.

PR #385, opened 2025-07-24
C Sharp | Initial build setup with build script

isaurab007:
*Issue #, if available:*

*Description of changes:* Native schema registry | C Sharp | Initial build setup with a build script to run all the Java/C and  C Sharp layers and test them for easier test automation.


native-schema-registry/build-csharp-native-schema-registry.sh
```diff
@@ -0,0 +1,51 @@
...
+cd ../..
+mvn install -P native-image
+echo
+
+echo "=== Step 5: Second C Library Build ==="
+cd c
+cmake -S. -Bbuild 
+cd build 
+cmake --build .
+echo
```
knoxy5467: We should not be playing hopscotch with builds. When we run a build command we need to have it either succeed or fail. This cyclic dependency is not intuitive for anybody using this package.

native-schema-registry/build-csharp-native-schema-registry.sh
```diff
@@ -0,0 +1,51 @@
...
+echo "=== Step 2: First Native Schema Registry Build ==="
+cd native-schema-registry
+mvn install -P native-image
+echo
+
+echo "=== Step 3: First C Library Build ==="
+cd c
+cmake -S. -Bbuild 
+cd build 
+cmake --build .
```
knoxy5467: this cmake is not needed here.

knoxy5467 (changes requested):

PR #397, opened 2025-08-12
Update native-image configuration to enable additional http URL protocol to hit IMDS + Remove unused native build arguments

isaurab007:
*Description of changes:*

Update native-image configuration to enable additional http URL protocol to hit IMDS + Remove unused native build arguments. This will make sure the compiled native library can pick up EC2 instance roles

### DIFF
--- a/native-schema-registry/pom.xml
+++ b/native-schema-registry/pom.xml
@@ -151,6 +151,8 @@
                                     -H:CLibraryPath=${project.basedir}/c/build/src
                                     -H:ResourceConfigurationFiles=${project.basedir}/src/main/resources/resource-config.json
                                     -H:Name=libnativeschemaregistry.so
+                                    --enable-url-protocols=http,https
+                                    --enable-all-security-services
                                 </buildArgs>
                             </buildArgs>
                         </configuration>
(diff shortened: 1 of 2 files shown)


native-schema-registry/pom.xml
```diff
@@ -151,6 +151,8 @@
                                     -H:CLibraryPath=${project.basedir}/c/build/src
                                     -H:ResourceConfigurationFiles=${project.basedir}/src/main/resources/resource-config.json
                                     -H:Name=libnativeschemaregistry.so
+                                    --enable-url-protocols=http,https
+                                    --enable-all-security-services
```
blacktooth: I think this comes with additional dependencies. Is `https` and `--enable-all-security-services` absolutely needed?

isaurab007: Main change in focus here is to have http option so we can make requests to IMDS. I will verify if we need to have the https option. `security-services` option is for https to work, as it requires the Java Cryptography Architecture (JCA) framework.

isaurab007: I have verified that it works with just the `http` protocol option. Pushed a follow-up commit accordingly.

blacktooth (approved):

PR #401, opened 2025-08-13
Create a C# build workflow

thiagoavelino:
This workflow is going to be automatically triggered on push, pr and release requests. This build is set to install the required dependencies and build the C# package using maven. No integ tests are executed yet. This commit also includes a change on the datatypes class so that the header files can be automatically referenced at build time.


isaurab007 (approved):
LGTM

PR #403, opened 2025-08-13
Use KafkaFlow in-the box interfaces + Use nupkg to distribute the csharp library instead of path to the DLL + Parent version downgrade in native pom

isaurab007:
**Issue #, if available:**

Currently, the way to use the library is:
```
  <ItemGroup>
    <Reference Include="AWSGsrSerDe">
      <HintPath>../../csharp/AWSGsrSerDe/AWSGsrSerDe/bin/Release/net8.0/AWSGsrSerDe.dll</HintPath>
    </Reference>
  </ItemGroup>
```
which is not user-friendly (expects the library to be on the machine and doesn't support indices for sharing nukpg).

Moving forward, it will be

``` 
<ItemGroup>
    <PackageReference Include="AWSGsrSerDe" Version="1.1.10" />
 </ItemGroup>
```

and the local index can be configured in `nuget.config` as

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="local-packages" value="../csharp/AWSGsrSerDe/AWSGsrSerDe/bin/Release" />
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
  </packageSources>
</configuration>
```

**Kafka Flow interfaces change**
Earlier we were using KafkaFlow interfaces within the demo project. Now, we shifted to using the KafkaFlow interfaces which come in-the-box in our csharp library.

**Pom parent version downgrade**
`1.1.24` produces `getCType` errors when running protobuf tests. This PR reverts this to `1.1.10` until we fix this issue in a follow up PR. 

**Description of changes:**
- Use KafkaFlow in-the box interfaces
- Use nupkg to distribute the csharp library instead of path to the DLL 
- Parent version downgrade in native pom

**Csharp tests work**
image

**Demo project works:**
image


native-schema-registry/ProtobufGSRKafkaDemo/Consumer/Consumer.csproj
```diff
@@ -18,4 +18,8 @@
     <ProjectReference Include="../Shared/Shared.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="AWSGsrSerDe" Version="1.1.10" />
```
blacktooth: TODO: Eventually, we want to use a placeholder here for the version.

native-schema-registry/ProtobufGSRKafkaDemo/Consumer/Program.cs
```diff
@@ -2,8 +2,9 @@
 using Microsoft.Extensions.DependencyInjection;
```
blacktooth, line 2: We can move this demo into a separate "demos/csharp/" directory

native-schema-registry/ProtobufGSRKafkaDemo/Consumer/Consumer.csproj
isaurab007: Done.

native-schema-registry/ProtobufGSRKafkaDemo/Consumer/Program.cs
isaurab007: Done.

blacktooth (approved):

PR #405, opened 2025-08-16
Onboard Musl variant of the native library 

isaurab007:
This PR:

1. Creates a maven build profile for musl java native library compilation
2. Onboards a Dockerfile showcasing an E2E C# run on Alpine Linux (which comes with musl libc)
3. Includes a musl build README showcasing steps to compile java using graalvm/musl and proceed run C# tests

Dockerfile builds all the way up-to C# layer and the tests work fine (except getCType errors which we are solving in a follow up PR)
image

While the current solution works, there are ToDos that need to be fixed as follow up PRs:
1. Permanently update library to 1.1.24 version across all modules and solve getCType errors
2. Fix C tests that do not work with musl as cmocka is not musl libc compatible

### DIFF
--- a/native-schema-registry/build/musl/Dockerfile
+++ b/native-schema-registry/build/musl/Dockerfile
@@ -1,0 +1,98 @@
+# ==========================
+# Stage 1 — Java build (GraalVM musl)
+# ==========================
+FROM ghcr.io/graalvm/native-image-community:17-muslib AS java-build
+
+WORKDIR /tmp
+
+# Build PIC-compatible zlib to fix shared library linking
+RUN curl -L https://github.com/madler/zlib/releases/download/v1.3/zlib-1.3.tar.gz | tar -xz && \
+    cd zlib-1.3 && \
+    CC=x86_64-linux-musl-gcc CFLAGS="-fPIC -O2" ./configure --static && \
+    make && \
+    cp libz.a /usr/local/musl/lib/libz.a
+
+# Copy the project
+COPY . .
+
+# Install packages for Java build
+RUN microdnf install -y maven git unzip zip curl bash findutils ca-certificates tar gzip \
+ && microdnf clean all
+
+# Install packages for C build
+RUN microdnf install -y dnf dnf-plugins-core && \
+    dnf -y config-manager --set-enabled ol9_codeready_builder && \
+    dnf -y install oracle-epel-release-el9 && \
+    dnf -y --enablerepo=ol9_developer_EPEL install swig lcov gcc clang clang-tools-extra make cmake && \
+    dnf clean all && rm -rf /var/cache/dnf
+
+# Build core Java library
+RUN mvn -U clean install -Dcheckstyle.skip=true -DskipTests
+RUN pwd
+# Build  C native libraries
+WORKDIR /tmp/native-schema-registry/c
+RUN cmake -S . -B build \
+  -DCMAKE_C_COMPILER=x86_64-linux-musl-gcc \
+  -DCMAKE_POSITION_INDEPENDENT_CODE=ON 
+WORKDIR /tmp/native-schema-registry/c/build
+RUN cmake --build . || true
+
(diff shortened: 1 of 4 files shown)


native-schema-registry/pom.xml
```diff
@@ -30,7 +30,7 @@
     <parent>
         <groupId>software.amazon.glue</groupId>
         <artifactId>schema-registry-parent</artifactId>
-        <version>1.1.10</version>
+        <version>1.1.24</version>
```
knoxy5467, line 33: `getCtype` issues still persists with this. Keep this at 1.1.10 until we can resolve the protobuf Ctype issues.

native-schema-registry/c/src/CMakeLists.txt
```diff
@@ -27,15 +27,16 @@ set_target_properties(${DATA_TYPES_MODULE_NAME} PROPERTIES
...
     INSTALL_RPATH "$ORIGIN"
 )
-add_custom_command(
-        TARGET ${DATA_TYPES_MODULE_NAME}
-        POST_BUILD
-        COMMENT "copying data types for maven project build"
-        COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:${DATA_TYPES_MODULE_NAME}> ${LIB_NATIVE_SCHEMA_REGISTRY_PATH}/
-        COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_LINKER_FILE:${DATA_TYPES_MODULE_NAME}> ${LIB_NATIVE_SCHEMA_REGISTRY_PATH}/
-        COMMENT "building maven project, native image target"
-        COMMAND mvn clean install -P native-image WORKING_DIRECTORY ${MAVEN_PROJECT_DIR}/..
```
knoxy5467, line 37: instead of removing this lets have the java .so artifacts pulled back out onto the local system. We can change this post build command to run the docker build.

native-schema-registry/pom.xml
```diff
@@ -159,5 +159,41 @@
...
+                                    -H:Name=libnativeschemaregistry.so
+                                    --libc=musl
+                                    --enable-url-protocols=http
+                                </buildArgs>
+                            </buildArgs>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
```
knoxy5467, line 197: We should have profiles / targets for musl build and for glibc build as we will be needing to target both in the future. Glibc is used by a majority of operating systems.

native-schema-registry/c/src/CMakeLists.txt
knoxy5467:
```
# Stage 1: Build the Maven project
FROM maven:3.8.6-jdk-11 AS builder
WORKDIR /app
COPY pom.xml .
COPY src ./src
RUN mvn clean install -DskipTests

# Stage 2: Create the final image
FROM openjdk:11-jre-slim
WORKDIR /app
COPY --from=builder /app/target/*.jar app.jar
ENTRYPOINT ["java", "-jar", "app.jar"]
```

Example

native-schema-registry/build/musl/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+    && chmod +x grpc_csharp_plugin
+
+# RUN ls -l /root/.nuget/packages/grpc.tools/2.48.1/tools/linux_x64/ && \
+#     file /root/.nuget/packages/grpc.tools/2.48.1/tools/linux_x64/protoc || true && \
+#     ldd /root/.nuget/packages/grpc.tools/2.48.1/tools/linux_x64/protoc || true
+
+# need to build it twice due to protobuf compilation misorder (normal for csharp builds today)
+RUN dotnet build . || true 
+RUN dotnet build . && dotnet test . || true
+
```
knoxy5467, line 100: Stage 2 is not needed. We should rely on the cmake build system that we have already built out instead of hiding the cyclic build system through a docker build.

PR #409, opened 2025-08-19
Add low level golang interaction with c swig wrapper.

knoxy5467:
This adds the pkg directory files to the golang branch.

These files are the low level interactions with the swig bindings.

### DIFF
--- a/native-schema-registry/golang/pkg/gsrserde-go/deserializer.go
+++ b/native-schema-registry/golang/pkg/gsrserde-go/deserializer.go
@@ -1,0 +1,171 @@
+package gsrserde
+
+import (
+	"runtime"
+
+	"github.com/awslabs/aws-glue-schema-registry/native-schema-registry/golang/pkg/GsrSerDe"
+)
+
+// Deserializer is a wrapper around the native schema registry deserializer
+// NOTE: This wrapper is NOT thread-safe. Each instance should be used by
+// only one context/operation to comply with native library constraints.
+type Deserializer struct {
+	deserializer GsrSerDe.Glue_schema_registry_deserializer
+	closed       bool
+}
+
+// NewDeserializer creates a new deserializer instance
+func NewDeserializer(configPath string) (*Deserializer, error) {
+	runtime.LockOSThread()
+	err := createErrorHolder()
+	
+	// Create native deserializer
+	deserializer := GsrSerDe.NewGlue_schema_registry_deserializer(configPath,err)
+	
+	if err != nil && err.Swigcptr() != 0 {
+		return nil, extractError("create deserializer", err)
+	}
+	
+	if deserializer == nil || deserializer.Swigcptr() == 0 {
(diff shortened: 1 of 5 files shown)


native-schema-registry/golang/pkg/gsrserde-go/deserializer.go
```diff
@@ -0,0 +1,171 @@
...
+	d := &Deserializer{
+		deserializer: deserializer,
+		closed:       false,
+	}
+	
+	return d, nil
+}
+
+// Decode deserializes the encoded data
+func (d *Deserializer) Decode(data []byte) ([]byte, error) {
```
thiagoavelino, line 42: Should we call this function deserialize?

native-schema-registry/golang/pkg/gsrserde-go/serializer.go
```diff
@@ -0,0 +1,114 @@
...
+// Serializer is a wrapper around the native schema registry serializer
+// NOTE: This wrapper is NOT thread-safe. Each instance should be used by
+// only one context/operation to comply with native library constraints.
+type Serializer struct {
+	serializer GsrSerDe.Glue_schema_registry_serializer
+	closed     bool
+}
+
+// NewSerializer creates a new serializer instance
+func NewSerializer(configPath string) (*Serializer, error) {
```
thiagoavelino, line 18: Is it the plan to allow pass the credential provider and schema version id like we do in Java?
https://github.com/awslabs/aws-glue-schema-registry/blob/master/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/GlueSchemaRegistryKafkaSerializer.java#L49-L57

native-schema-registry/golang/pkg/gsrserde-go/deserializer.go
thiagoavelino: Are we matching the java versions?

native-schema-registry/golang/pkg/gsrserde-go/deserializer.go
```diff
@@ -0,0 +1,171 @@
...
+	err := createErrorHolder()
+	
+	// Create native deserializer
+	deserializer := GsrSerDe.NewGlue_schema_registry_deserializer(configPath,err)
+	
+	if err != nil && err.Swigcptr() != 0 {
+		return nil, extractError("create deserializer", err)
+	}
+	
+	if deserializer == nil || deserializer.Swigcptr() == 0 {
```
isaurab007, line 29: `deserializer.Swigcptr() == 0`

Can you explain why we return `ErrInitializationFailed` in this case?

thiagoavelino (approved):

isaurab007 (approved):
LGTM. Discussed we will review unit tests as a part of PR that has user facing interfaces.

rohanmurthy (approved):

PR #417, opened 2025-08-21
Create a C# build workflow. This workflow is going to be automaticall…

thiagoavelino:
*Description of changes:*

This workflow build is set to install the require dependencies and build the C# package using maven. No integ tests are executed yet. This commit also includes a change on the datatypes class so that the header files can be automatically referenced at build time.


isaurab007:
The build is failing as native-schema-registry is trying to read 1.1.10 version of parent pom (which would be fetched from maven and not .m2 given we are building 1.1.24 of the parent). So technically, it should not fail in Java build step. 

What is happening is that the 1.1.10 version in repo has some missing files that are a part of the java base added as a part of this commit: 
https://github.com/blacktooth/aws-glue-schema-registry/commit/fb4013c00905a95d15e7b0ab84993e5058aab6c2#diff-6c20d1c397b9e361c2de90f34e985a4f80d21bef1838502aa62a4d1b3600b0be

If we want to proceed with 1.1.10 as a medium term workaround, we will need to address this drift first.

PR #440, opened 2025-09-10
fix(kafkaconnect): resolve secondary deserializer schema extraction for non-GSR data

rohanmurthy:
### Problem
The `AWSKafkaAvroConverter` was incorrectly attempting to extract schemas from all data using GSR metadata, even for data that should be handled by secondary deserializers (e.g., Confluent Schema Registry data). This caused failures when processing non-GSR data with the error:

```
GlueSchemaRegistryIncompatibleDataException: Data is not compatible with schema registry size: X
```

### Root Cause
The `toConnectData()` method unconditionally called `getSchemaDefinition()` on the raw byte array, bypassing the secondary deserializer routing logic that correctly handled deserialization but not schema extraction.

### Solution
- Add conditional logic to detect data type using `canDeserialize()`
- Route schema extraction based on data source:
  - **GSR data**: Extract schema from registry metadata (existing behavior)
  - **Secondary deserializer data**: Extract schema from deserialized Avro object
- Maintain full backward compatibility for existing GSR users

### Changes
- **Modified**: `AWSKafkaAvroConverter.toConnectData()` - Updated to use conditional schema extraction
- **Added**: `extractAvroSchema()` - Main routing logic for schema extraction
- **Added**: `extractSchemaFromAvroObject()` - Extracts schema from GenericRecord/SpecificRecord
- **Added**: Comprehensive unit tests covering both GSR and secondary deserializer paths
- **Added**: Integration tests with real Avro objects

### Testing
- ✅ All existing tests pass (backward compatibility verified)
- ✅ New tests cover GSR data path, secondary deserializer path, and error scenarios
- ✅ Integration tests verify end-to-end functionality

### Breaking Changes
None - this is a bug fix that maintains full backward compatibility.

Fixes secondary deserializer functionality for Kafka Connect users migrating from other schema registries.

### DIFF
--- a/avro-kafkaconnect-converter/src/test/java/com/amazonaws/services/schemaregistry/kafkaconnect/AWSKafkaAvroConverterTest.java
+++ b/avro-kafkaconnect-converter/src/test/java/com/amazonaws/services/schemaregistry/kafkaconnect/AWSKafkaAvroConverterTest.java
@@ -227,6 +227,35 @@ public class AWSKafkaAvroConverterTest {
     }
 
     /**
+     * Test that the fix for secondary deserializer schema extraction works with GSR data.
+     * This ensures backward compatibility is maintained.
+     */
+    @Test
+    void testConverter_toConnectData_GSRData_BackwardCompatibility() {
+        Struct expected = createStructRecord();
+        String avroSchemaDefinition = avroData.fromConnectSchema(expected.schema()).toString();
+        Object avroData = this.avroData.fromConnectData(expected.schema(), expected);
+
+        AWSKafkaAvroSerializer awsKafkaAvroSerializer = createSerializer(avroSchemaDefinition, schemaVersionIdForTesting);
+        AWSKafkaAvroDeserializer awsKafkaAvroDeserializer = createDeserializer(avroData, genericBytes, avroSchemaDefinition);
+        
+        // Mock canDeserialize to return true for GSR data
+        when(awsKafkaAvroDeserializer.getGlueSchemaRegistryDeserializationFacade().canDeserialize(genericBytes))
+                .thenReturn(true);
+        
+        converter = new AWSKafkaAvroConverter(awsKafkaAvroSerializer, awsKafkaAvroDeserializer, this.avroData);
+
+        byte[] serializedData = converter.fromConnectData(testTopic, expected.schema(), expected);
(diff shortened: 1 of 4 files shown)


avro-kafkaconnect-converter/src/test/java/com/amazonaws/services/schemaregistry/kafkaconnect/AWSKafkaAvroConverterTest.java
```diff
@@ -266,7 +295,7 @@ private AWSKafkaAvroDeserializer createDeserializer(Object record,
                 .build();
 
         when(glueSchemaRegistryDeserializationFacade.deserialize(awsDeserializerInput)).thenReturn(record);
-        when(glueSchemaRegistryDeserializationFacade.getSchemaDefinition(bytes)).thenReturn(schemaDefinition);
+        when(glueSchemaRegistryDeserializationFacade.getSchemaDefinition(eq(bytes))).thenReturn(schemaDefinition);
```
blacktooth, line 298: Why do we have to change this?

blacktooth (approved):

avro-kafkaconnect-converter/src/test/java/com/amazonaws/services/schemaregistry/kafkaconnect/AWSKafkaAvroConverterTest.java
knoxy5467: @rohanmurthy @blacktooth can we resolve this so we can determine if we are going to merge this?

rohanmurthy: My change adds a new test (`testConverter_toConnectData_GSRData_BackwardCompatibility`) that mocks the same method (`getSchemaDefinition()`) using the `eq()` argument matcher. When multiple mock setups exist for the same method, Mockito requires ALL setups to use explicit argument matchers. You cannot mix raw parameters with matchers. Since the new test uses `eq(genericBytes)`, this existing setup must also use `eq(bytes)` instead of the raw bytes parameter.

`eq(bytes)` makes it clear that we're doing exact byte array equality matching, improving test readability and maintainability. If additional mock setups are added later with different matchers (like `any()`, `argThat()`, etc.), this setup won't break due to matcher inconsistency.

PR #442, opened 2025-09-11
GraalVM-17-to-21-update

knoxy5467:
Updating readme's and musl build scripts to reference building with graalvm-ce-21 instead of 17.


native-schema-registry/musl-build-helper-scripts/musl/Dockerfile.graalvm
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/graalvm/native-image-community:17-muslib
+FROM ghcr.io/graalvm/native-image-community:21-muslib
```
isaurab007, line 1: Have a verified that musl build works fine with this image? Just confirming again.

knoxy5467: yes. Have done it many times as a part of troubleshooting.

PR #459, opened 2025-09-19
Upgarde csharp library version to OneOneTwentyFive to match java base we are using + Make demo producers multi-threaded

isaurab007:
# Issue #, if available:
1. The csharp library is versioned @ 1.1.10 which lost parity with Java base.
2. The producers in the demo use only one thread and therefore are not production-like examples.

# Description of changes:
1. This PR restores that parity of versioning in c layer and java base
2. Add multi-threaded producers making it production-like.


isaurab007:
Yes, all the tests pass. 

image

PR #460, opened 2025-09-19
Native caching protobuf

knoxy5467:
This pull request adds a caching mechanism for protobuf schema file descriptors in the `ProtobufPreprocessor` class to improve performance and reduce redundant parsing. It introduces a thread-safe, configurable cache using Guava's `LoadingCache`, and includes comprehensive unit tests to verify cache behavior, eviction, error handling, and thread safety.

### Caching and performance improvements

* Added a static, thread-safe `LoadingCache` to `ProtobufPreprocessor` for mapping schema definitions to `FileDescriptor` objects, reducing repeated parsing and improving performance. The cache size is configurable and defaults to 200 entries. [[1]](diffhunk://#diff-23c5f26757077053e654439751e0d464e7e01e0e1e063d1d8a37a24eaa2bb9b1R6-R49) [[2]](diffhunk://#diff-23c5f26757077053e654439751e0d464e7e01e0e1e063d1d8a37a24eaa2bb9b1R59-R104)
* Modified `prefixMessageIndexToBytes` to retrieve file descriptors from the cache, initializing the cache on first use if necessary. [[1]](diffhunk://#diff-23c5f26757077053e654439751e0d464e7e01e0e1e063d1d8a37a24eaa2bb9b1R6-R49) [[2]](diffhunk://#diff-23c5f26757077053e654439751e0d464e7e01e0e1e063d1d8a37a24eaa2bb9b1R59-R104)
* Updated `initializeSerializerWithConfig` to initialize the cache with the configured size during serializer setup.

### Testing and reliability

* Added a new test class `ProtobufPreprocessorTest` with unit tests covering cache hits, eviction, zero-size cache, invalid schema handling, custom cache size, and thread safety for cache initialization.

### Thread safety and configuration

* Ensured thread-safe cache initialization using double-checked locking and provided a synchronized method to explicitly set cache size.

These changes collectively improve the efficiency, reliability, and configurability of protobuf schema handling in the schema registry serialization process.

### DIFF
--- a/native-schema-registry/src/main/java/com/amazonaws/services/schemaregistry/GlueSchemaRegistrySerializationHandler.java
+++ b/native-schema-registry/src/main/java/com/amazonaws/services/schemaregistry/GlueSchemaRegistrySerializationHandler.java
@@ -67,6 +67,7 @@ public class GlueSchemaRegistrySerializationHandler {
             String filePath = CTypeConversion.toJavaString(configFilePath);
             Map<String, String> configs = ConfigurationFileReader.loadConfigFromFile(filePath);
             NativeGlueSchemaRegistryConfiguration configuration = new NativeGlueSchemaRegistryConfiguration(configs);
+            ProtobufPreprocessor.initializeCache(configuration.getCacheSize());
             SerializerInstance.create(configuration);
             return 0;
         } catch (Exception e) {


native-schema-registry/src/main/java/com/amazonaws/services/schemaregistry/serializer/ProtobufPreprocessor.java
```diff
@@ -35,4 +56,50 @@ public static byte[] prefixMessageIndexToBytes(byte[] bytesToEncode, String sche
...
+    private static Descriptors.FileDescriptor getFromCache(@Nonnull String schemaDef) throws UncheckedExecutionException {
+
+        if (schemaDefToFileDescriptorCache == null) {
+            synchronized (ProtobufPreprocessor.class) {
+                if (schemaDefToFileDescriptorCache == null) {
+                    initializeCache(cacheLimit);
+                }
+            }
+        }
+        return schemaDefToFileDescriptorCache.getUnchecked(schemaDef);
```
isaurab007, line 83: Can you document the reason behind going with `getUnchecked` over `get` as a line comment?

isaurab007: Also, why does `initializeCache` need to be thread safe but not `getUnchecked`? Is it already thread safe?

native-schema-registry/test/java/com/amazonaws/services/schemaregistry/serializer/ProtobufPreprocessorTest.java
```diff
@@ -0,0 +1,249 @@
...
+        ProtobufPreprocessor.prefixMessageIndexToBytes(
+                testBytes, basicSyntax2Schema,
+                "com.amazonaws.services.schemaregistry.tests.protobuf.syntax2.basic.Phone");
+
+        assertEquals(missCountBefore + 1, cache.stats().missCount(),
+                "Should have one more miss");
+    }
+
+    @Test
+    public void testThreadSafetyOfCacheInitialization() throws Exception {
```
isaurab007, line 183: Same thing, do we want to add another test for parallel cache access (getUnchecked) here? Or does `google.common.cache` library make any claims about thread safety of the method?

native-schema-registry/src/main/java/com/amazonaws/services/schemaregistry/serializer/ProtobufPreprocessor.java
isaurab007, line 83: Have we tested scenarios where the size of `schemaDef` is huge (multiple KBs or even larger)? Does it have any impact on memory? Do we want to use its hash as the key instead?

native-schema-registry/test/java/com/amazonaws/services/schemaregistry/serializer/ProtobufPreprocessorTest.java
```diff
@@ -0,0 +1,249 @@
+package com.amazonaws.services.schemaregistry.serializer;
```
isaurab007, line 1: Nit: I think we should also write tests for the functionality of `prefixMessageIndexToBytes` and `convertBase64SchemaToStringSchema` apart from caching, so we can validate the main methods in this class. Can go in as another PR.

isaurab007 (changes requested):

native-schema-registry/test/java/com/amazonaws/services/schemaregistry/serializer/ProtobufPreprocessorTest.java
knoxy5467: We are using this cache specifically because it is thread safe.

native-schema-registry/test/java/com/amazonaws/services/schemaregistry/serializer/ProtobufPreprocessorTest.java
knoxy5467: Those tests can be added at a later date. We should not add extra tests for testing functionality that was not added from this PR.

native-schema-registry/src/main/java/com/amazonaws/services/schemaregistry/serializer/ProtobufPreprocessor.java
knoxy5467: if you check the documentation: https://guava.dev/releases/11.0.2/api/docs/com/google/common/cache/CacheBuilder.html

you will find that the cache uses hashing for keys.

Additionally this cache uses the same configuration value at startup as the same cache we use in the serde.

knoxy5467: getUnchecked is already thread safe as the cache itself is thread safe. 

the initialization is not thread safe as we do not want multiple threads to attempt to init the same object.

Additionally get vs get unchecked is just a difference in the type of exception that is expected from the underlying cache. The .parse() method throws and unchecked exception class.

native-schema-registry/src/main/java/com/amazonaws/services/schemaregistry/serializer/ProtobufPreprocessor.java
```diff
@@ -3,29 +3,50 @@
...
+import java.util.concurrent.ExecutionException;
+
+import javax.annotation.Nonnull;
 
 public class ProtobufPreprocessor {
     private static final ProtobufWireFormatEncoder ENCODER = new ProtobufWireFormatEncoder(new MessageIndexFinder());
     private static final MessageIndexFinder MESSAGE_INDEX_FINDER = new MessageIndexFinder();
 
+    private static volatile long cacheLimit = 200; // same default as GlueSchemaRegistryConfiguration
+    private static volatile LoadingCache<String, Descriptors.FileDescriptor> schemaDefToFileDescriptorCache;
```
blacktooth, line 30: Can we move the cache deeper into Java layer?

knoxy5467: What do you mean as to moving this deeper into the Java layer?

we already do schema caching at the next layer with the calls to AWS.

PR #464, opened 2025-09-24
Move away from dataFormat config in C# by passing it manually in Serialize method

isaurab007:
# Issue #, if available:
Today, other languages do not rely on dataFormat attribute in the configuration file and do not need them at the time of serializer initialization. However, this pattern is inconsistent with Csharp.

# Description of changes:
This change does away the necessity of the `dataFormat` in the config file making it similar to golang and java.

# New usage
```csharp
private static readonly string CONFIG_PATH = "<PATH_TO_CONFIG_FILE>";
var protobufSerializer = new GlueSchemaRegistryKafkaSerializer(CONFIG_PATH);
var serialized = protobufSerializer.Serialize(message, message.Descriptor.FullName, "PROTOBUF");
// send serialized bytes to Kafka using producer.Produce(serialized)
```

# Test results after the changes
image


native-schema-registry/csharp/AWSGsrSerDe/AWSGsrSerDe.Tests/deserializer/GlueSchemaRegistryKafkaDeserializerTests.cs
```diff
@@ -31,8 +31,10 @@ namespace AWSGsrSerDe.Tests.deserializer
     [TestFixture]
     public class GlueSchemaRegistryKafkaDeserializerTests
     {
-        private static readonly string AVRO_CONFIG_PATH = GetConfigPath("configuration/test-configs/valid-minimal.properties");
-        private static readonly string PROTOBUF_CONFIG_PATH = GetConfigPath("configuration/test-configs/valid-minimal-protobuf.properties");
+        private static readonly string CONFIG_PATH = GetConfigPath("configuration/test-configs/valid-minimal.properties");
+        private const string AVRO_DATA_FORMAT = "AVRO";
+        private const string JSON_DATA_FORMAT = "JSON";
+        private const string PROTOBUF_DATA_FORMAT = "PROTOBUF";
```
knoxy5467, line 37: these constants should probably be declared in the actual library itself instead of just in the tests.

native-schema-registry/csharp/AWSGsrSerDe/AWSGsrSerDe/serializer/GlueSchemaRegistryKafkaSerializer.cs
```diff
@@ -34,9 +33,7 @@ public class GlueSchemaRegistryKafkaSerializer
         /// <param name="configFilePath">Path to the configuration properties file</param>
         public GlueSchemaRegistryKafkaSerializer(string configFilePath)
         {
-            _dataFormat = ConfigFileReader.GetConfigValue(configFilePath, "dataFormat") ?? "AVRO";
             _schemaNamingStrategy = new DefaultSchemaNameStrategy();
-            
             _glueSchemaRegistrySerializer = new GlueSchemaRegistrySerializer(configFilePath);
         }
```
knoxy5467, line 38: to stay consistent with java https://github.com/knoxy5467/aws-glue-schema-registry/blob/9a9d6cd5e151084d0c772466299bccdf152d9faf/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/GlueSchemaRegistryKafkaDeserializer.java#L68-L74 We should probably keep the dataformat as a part of the csharp initialization. Just instead of it being pulled from the config file it should be passed in.

native-schema-registry/csharp/AWSGsrSerDe/AWSGsrSerDe/serializer/GlueSchemaRegistryKafkaSerializer.cs
```diff
@@ -47,22 +44,22 @@ public GlueSchemaRegistryKafkaSerializer(string configFilePath)
...
         /// <returns>serialized byte array</returns>
-        public byte[] Serialize(object data, string topic)
+        public byte[] Serialize(object data, string topic, string dataFormat)
         {
             if (data == null)
             {
                 return null;
             }
 
-            var serializer = DataFormatSerializerFactory.GetInstance().GetSerializer(_dataFormat);
```
knoxy5467, line 57: As I mentioned in earlier comments. I think that this method was fine. Simply where we get data format from should probably be changed.

If we want to break from java and have the data format serde be declared at serialize time then that is fine too. But should be documented as a split from java.

isaurab007:
We thought more about this and seems like an extra dataformat config at initialization feels redundant given we are already passing in a config file path as the very adjacent argument.

knoxy5467:
The decision was not about it being redundant but about moving this to be re-evaluated at a later date.

PR #469, opened 2025-09-29
Write a SerDes lifecycle test for memory leak validation using ASan

isaurab007:
## Issue #, if available:
Right now, although ASan is integrated with our build system, there is no test for the SerDes lifecycle (initiate, encode/dispose, dispose) and multiple encode/decode iterations in the C layer. Hence, there is no way to catch memory leaks in this flow.

## Description of changes:
This PR adds these tests where it covers:

1. test_serializer_lifecycle_memory_management
2. test_deserializer_lifecycle_memory_management
3. test_full_serialization_deserialization_cycle
4. test_multiple_operations_memory_management

### DIFF
--- a/native-schema-registry/c/CMakeLists.txt
+++ b/native-schema-registry/c/CMakeLists.txt
@@ -70,6 +70,8 @@ set(TEST_TARGETS
     glue_schema_registry_error_test
     glue_schema_registry_serializer_test
     glue_schema_registry_deserializer_test
+    # There are memory leaks in graal_create_isolate. Uncomment after fixing.
+    # schema_registry_memory_leak_test
 )
 
 # Manual quality gates target (always available)
--- a/native-schema-registry/c/test/CMakeLists.txt
+++ b/native-schema-registry/c/test/CMakeLists.txt
@@ -23,6 +23,8 @@ list(
         read_only_byte_array_test
         mutable_byte_array_test
         glue_schema_registry_error_test
+        # There are memory leaks in graal_create_isolate. Uncomment after fixing.
+        # schema_registry_memory_leak_test
 )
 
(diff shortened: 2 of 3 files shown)


native-schema-registry/c/CMakeLists.txt
```diff
@@ -70,6 +70,8 @@ set(TEST_TARGETS
     glue_schema_registry_error_test
     glue_schema_registry_serializer_test
     glue_schema_registry_deserializer_test
+    # There are memory leaks in graal_create_isolate. Uncomment after fixing.
+    # schema_registry_memory_leak_test
```
knoxy5467, line 74: Can this be more descriptive as to where and what the memory leak is from?

native-schema-registry/c/test/CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ list(
         read_only_byte_array_test
         mutable_byte_array_test
         glue_schema_registry_error_test
+        # There are memory leaks in graal_create_isolate. Uncomment after fixing.
```
knoxy5467, line 26: graal_create is outside of our projects scope correct? Additionally if there are leaks in graalvm then we should raise and issue on their repo.

native-schema-registry/c/test/CMakeLists.txt
```diff
@@ -67,4 +70,4 @@ if (CMAKE_SYSTEM_NAME MATCHES "^(Linux)$")
     include(cmake/CodeCoverage.cmake)
     set(min_code_coverage 100)
     setup_target_for_coverage(coverage "ctest ." coverage ${min_code_coverage})
-endif()
```
knoxy5467, line 70: NIT: unneeded formatting changes. Only fix if second revision is needed.

PR #494, opened 2025-12-17
fix(validation): Add schema compatibility check to RegisterSchemaVersion for JSON data format

rohanmurthy:
TODO

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -98,6 +98,10 @@
             <artifactId>guava</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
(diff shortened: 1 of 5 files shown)


jvdadda:
Hi @rohanmurthy, this addressed a real gap: Glue enforces the compatibility mode for Avro and Protobuf but not for JSON, so a `BACKWARD`-breaking JSON version registers silently and only fails later, in a consumer. We ported the check into our maintained fork of the Java library in https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/119. It is available in the latest published release, currently 4.0.0, on Maven Central as `com.mobsuccess:schema-registry-serde`.

Two deliberate deviations, for transparency. It is opt-in (`jsonSchemaCompatibilityCheckEnabled`, default `false`): a client-side check that can refuse registrations, plus one extra `GetSchemaVersion` call per new definition, felt too big to switch on for every JSON producer as a side effect of upgrading. And the mode is read from the enum rather than from the prefix of its name, so `DISABLED` skips the check like `NONE` does. What it compares (the `required` contract) is documented at https://github.com/mobsuccess-devops/aws-glue-schema-registry/blob/master/docs/configuration.md#limits-of-the-json-compatibility-check.

PR #516, opened 2026-03-12
[Java] Remove unused jimfs dependency

### LINKED ISSUE #353: Unnecessary dependency on jimfs?
pom.xml mentions a dependency on com.google.jimfs:jimfs, but it appears to be unused (= a search on Github turns up no references in the code). It looks like some uses of jimfs were removed in https://github.com/awslabs/aws-glue-schema-registry/pull/230. Can the dependency itself be removed as well?

jvdadda:
*Issue #, if available:* Fixes #353

*Description of changes:*

Remove the unused `com.google.jimfs:jimfs` direct dependency from `serializer-deserializer/pom.xml`.

This dependency was left behind after its usages were removed in #230. No code in the project imports any jimfs classes. The library remains available as a transitive dependency (via `wire-compiler`) for the `multilang-schema-registry` GraalVM native-image build, so the `native-image.properties` configuration continues to work.

### DIFF
--- a/serializer-deserializer/pom.xml
+++ b/serializer-deserializer/pom.xml
@@ -189,11 +189,6 @@
             <artifactId>proto-google-common-protos</artifactId>
             <version>2.7.4</version>
         </dependency>
-        <dependency>
-            <groupId>com.google.jimfs</groupId>
-            <artifactId>jimfs</artifactId>
-            <version>1.1</version>
-        </dependency>
         <!-- Used for Proto assertions in tests -->
         <dependency>
             <groupId>com.google.truth.extensions</groupId>


jvdadda:
Closing this one. This repository is not taking outside contributions in practice — this PR and nine others have sat without a single review since March — so the change ships in our fork instead, which has diverged too far anyway (Java only, Maven → Gradle, and a Java → Kotlin conversion) to keep proposing the patch here.

**Ported as** https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/71 — same removal, expressed as the Gradle dependency plus its version catalog entry.

Reopen it if anyone ever wants to take it.

Issue #350, opened 2024-07-08
New release with updated kafka.scala.version


vab2048:
The current release uses:         <kafka.scala.version>2.12</kafka.scala.version>

This is causing issues because of the binary compatibility breakage between 2.12 and 2.13. 

See this: https://github.com/spring-projects/spring-boot/issues/24621#issuecomment-753887474 for the issue I am facing.

We are unable to run embedded kafka without overhead in configuration because of this. 
Could you please bump the version to 2.13 and do a new release.

vab2048:
It would also be good to bump the avro version to the latest as well.

PR #310, opened 2023-11-14
Update the target scala version to 2.13 for mbknor-jackson-jsonschema and kafka

junyuc25:
*Issue #, if available:*

*Description of changes:*
Updating the following libraries to use the version compiled with Scala 2.13.
  1. mbknor-jackson-jsonschema_2.12 
  2. kafka_2.12

*Motivation:*
There is ongoing work to upgrade AWS SDK to V2 for the Apache Spark as tracked in [this Jira](https://issues.apache.org/jira/browse/SPARK-44124). When upgrading the Spark Kinesis connector module (which transitively depends on glue-schema-registry), the GSR library also pulls in `mbknor-jackson-jsonschema_2.12`. However, as Apache Spark 4.0 release will drop support for scala 2.12 ([Jira ticket](https://issues.apache.org/jira/browse/SPARK-44113)), this lib is considered as banned dependency, resulting in build failures. We would need to replace dependencies that are compiled with scala 2.12 within the GSR repository to proceed with the SDK upgrade. Hence this PR. 

Example build failure information:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.3.0:enforce (enforce-versions) on project spark-streaming-kinesis-asl_2.13:
[ERROR] Rule 2: org.apache.maven.enforcer.rules.dependency.BannedDependencies failed with message:
[ERROR] org.apache.spark:spark-streaming-kinesis-asl_2.13:jar:4.0.0-SNAPSHOT
[ERROR]   software.amazon.kinesis:amazon-kinesis-client:jar:2.5.2
[ERROR]     software.amazon.glue:schema-registry-serde:jar:1.1.14
[ERROR]       com.kjetland:mbknor-jackson-jsonschema_2.12:jar:1.0.39 <--- banned via the exclude/include list
```

*Testing:*
1. The unit tests passed with `mvn clean package`
2. Ran integration tests locally using `run-local-tests.sh`. Although there are 3 test failures, these failures are also observed when running against master branch. Hence they should not be caused by this PR.
```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   GlueSchemaRegistryKafkaIntegrationTest.testKafkaStreamsProcess:397->assertStreamsRecordsEquality:483 
Expected: is <0>
     but: was <6>
[ERROR]   GlueSchemaRegistryKafkaIntegrationTest.testKafkaStreamsProcess:397->assertStreamsRecordsEquality:483 
Expected: is <0>
[...]

PR #525, opened 2026-06-30
Move to Scala 2.13 (BREAKING: drops 2.12) - addresses #350

shakhihali:
## Description

Moves the library from Scala 2.12 to Scala 2.13 by setting
`kafka.scala.version` to `2.13` (and fixing a hardcoded `_2.12` suffix on the
`mbknor-jackson-jsonschema` dependency so it tracks the property).

Addresses #350.

## ⚠️ This is a breaking change — not a drop-in fix

Reviewer feedback correctly pointed out that this is the **same change as #310**,
which was merged and then reverted in **#314** ("This can potentially break the
builds"). This PR previously did not acknowledge that history. Flagging it
explicitly now:

- This repository publishes a **single set of artifacts**. Flipping
  `kafka.scala.version` does **not add** 2.13 support alongside 2.12 — it
  **swaps** the library onto the Scala 2.13 Kafka / mbknor jars and **drops
  2.12**.
- Every downstream consumer still on Scala 2.12 will break on the next release.
- The clean `dependency:tree` (only `_2.13` artifacts) is the *evidence of the
  drop*, not proof of safety. Likewise, this repo's integration tests pass
  because they run against the now-2.13 classpath; they do not exercise a
  downstream 2.12 application.

## Why do it anyway

Scala 2.13 is the correct long-term direction: Apache Kafka 4.0 and Apache
Spark 4.0 both drop Scala 2.12. See the original motivation in #310 (Spark 4.0
bans `_2.12` transitive dependencies).

## How this should ship

As a deliberate breaking change, owned by the maintainers:

1. **Major version bump** — proposed `2.0.0` (added to `CHANGELOG.md` in this PR).
2. **Release note** documenting that 2.12 support is dropped (added).
3. **Ideally, cross-published `_2.12` and `_2.13` artifacts** — this was the
   maintainer's stated preference on #314, and it would avoid stranding 2.12
   consumers. That requires build/release-pipeline changes (Scala-suffixed
   artifact variants, matrixed publish) that are a maintainer decision and are
   **not** included here.

## Status

Converted to **draft**. This should not merge as a silent version flip. I'm
leaving it open as the starting point for a deliberate 2.0.0 breaking release,
and deferring to the maintainers on whether to (a) take this flip as-is in a
major release, or (b) pursue cross-published artifacts instead. Happy to do the
cross-publishing build work if that's the preferred path.

## Testing

- `mvn clean verify` — BUILD SUCCESS across all 13 modules on the 2.13 classpath.
- Note (per the above): this validates the library builds and tests **on 2.13**;
  it does not and cannot demonstrate 2.12 consumers are unaffected — they are not.


knoxy5467:
Have you ran the integration tests for this change?

knoxy5467:
This is the same change as #310, which was reverted in #314 for breaking Scala 2.12 consumers. The diff is identical and nothing here addresses why it was backed out.

We publish a single set of artifacts. Flipping kafka.scala.version to 2.13 doesn't add 2.13 support — it swaps the library onto the 2.13 Kafka/mbknor jars and drops 2.12. Every 2.12 consumer breaks on the next release. The maintainer's call last time was to cross-publish _2.12 and _2.13, not flip the single build.

The dependency:tree showing only 2.13 artifacts is the breakage, not proof of safety — "no _2.12 anywhere" is exactly what strands a 2.12 consumer. And green integration tests here won't catch it either, since they run against this repo's now-2.13 classpath, not a downstream 2.12 app.

Moving to 2.13 is the right direction long-term — Kafka 4.0 and Spark 4.0 both drop 2.12. But that's a deliberate breaking change: major version bump, release note, ideally cross-published artifacts. Not a silent version flip. Handle it as a breaking change before merging.

shakhihali:
You're right on all counts, and I appreciate the detailed correction.

I confirmed the history: #310 made this exact flip (merged 2023-12-13) and #314 reverted it two days later ("This can potentially break the builds"). My PR was the same diff and didn't address why it was backed out. That was my miss.

The specific error in my reasoning: I treated the `dependency:tree` output showing only `_2.13` artifacts as proof of safety, when — as you point out — that *is* the regression. Since we publish a single artifact set, flipping `kafka.scala.version` swaps the library onto the 2.13 jars and drops 2.12 rather than adding it, so "no `_2.12` anywhere" is exactly what strands a 2.12 consumer. And the green integration tests here only prove the repo builds on its own now-2.13 classpath, not that a downstream 2.12 app survives. I shouldn't have presented it as a safe fix.

I've rescoped this rather than merge it as a silent flip:
- Converted the PR to **draft**.
- Added a **`CHANGELOG.md`** entry marking this a breaking change with a proposed **2.0.0** major bump.
- Rewrote the description to state plainly that this drops 2.12, why it's still the right long-term direction (Kafka 4.0 / Spark 4.0 both drop 2.12, per #310's original motivation), and that shipping it is a deliberate maintainer decision.

On the path forward — I'd defer to you between taking the flip as part of an intentional 2.0.0 release vs. cross-publishing `_2.12` and `_2.13` artifacts, which I understand was your stated preference on #314. If cross-publishing is the direction, I'm happy to do the build work (Scala-suffixed artifact variants + matrixed publish); just confirm that's what you'd like and I'll put up a separate PR for it.

subramp:
I rebased this onto current `master` and ran the full integration suites, Java and C#. Everything passes. Details below in case they are useful when this comes off draft.

**Important caveat up front:** these results say the flip is *mechanically* sound. They do not say dropping Scala 2.12 is safe for consumers. As noted above, this repo publishes a single artifact set, so green tests here run against the now-2.13 classpath and cannot exercise a downstream 2.12 application. The policy question is unchanged by anything below.

## Rebase

The branch is based on `4b9cac47`, and the `pom.xml` hunk conflicts with master. I resolved it by keeping master's `kafka.version` 3.9.2 (raised by #539 for CVE-2026-35554) and leaving out the `aws.sdk.v1.version` line that #538 removed, taking only `kafka.scala.version` 2.13 from here. Rebased onto `04dc718`, the diff is 3 files, +10/-2: the CHANGELOG entry plus the two pom lines, unchanged in substance.

## Java

`mvn --file integration-tests/pom.xml verify -Psurefire`, against a real Glue registry with Kafka and LocalStack in Docker:

```
Tests run: 74, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS

GlueSchemaRegistryKafkaIntegrationTest      45/45   729.4s
GlueSchemaRegistryKinesisIntegrationTest    28/28   497.3s
ProtobufSchemaConverterIntegrationTest       1/1      0.5s
```

Unit suites separately: **1974 tests across 11 modules, 1 failure**, and that one failure (`GlueSchemaRegistryConfigurationTest.testBuildConfig_noRegionConfigsSupplied_throwsException`) reproduces on clean `master` on the same host, because a region resolves from the environment and the expected exception never throws. Not related to this change.

## C#

Worth running because the flip reaches the C# binding: `multilang-schema-registry` depends on `schema-registry-serde`, so the Scala jars land in the GraalVM native image that the C layer and the C# SerDe wrap. I rebuilt the 140 MB `libnativeschemaregistry.so` from this branch and confirmed the new classpath, `mbknor-jackson-jsonschema_2.13-1.0.39.jar` and `scala-library-2.13.1.jar`, replacing the 2.12 pair.

```
dotnet test -c Release
Passed!  -  Failed: 0, Passed: 173, Skipped: 0, Total: 173, Duration: 5m 9s
```

## Resolution check

Before: `mbknor-jackson-jsonschema_2.12:1.0.39` → `scala-library:2.12.10`
After: `mbknor-jackson-jsonschema_2.13:1.0.39` → `scala-library:2.13.1`

Both suffixes publish at the same 1.0.39, so this is a pure suffix swap with no version movement. Also worth noting in the changelog: `kafka_2.12` stops at 3.9.2 on Maven Central while `kafka_2.13` continues through 4.3.1, so 3.9.2 is the last Kafka release available on 2.12 at all. That is a stronger argument than the Kafka 4.0 / Spark 4.0 framing currently in the entry.

Separately, the diff is complete. `pom.xml:85` and `serializer-deserializer/pom.xml:91` are the only hardcoded Scala 2.12 references in the repo; everything else already resolves through `${kafka.scala.version}`. The `serializer-deserializer` hunk fixes a real latent bug, since that module was ignoring the property the root pom already parameterizes.

## Two unrelated things I hit

Both pre-existing on `master`, flagging them only so nobody attributes them to this PR:

1. `multilang-schema-registry/c/src/CMakeLists.txt` invokes the Maven `native-image` goal itself, so CMake is the entry point and `JAVA_HOME` must point at GraalVM when running `cmake --build`. Running `mvn -P native-image-glibc` directly fails on `Invalid library path .../c/build/src`, and even after creating that directory it fails to link because `libnative_schema_registry_c_data_types` does not exist yet. `multilang-schema-registry/README.md` documents the opposite order.
2. CMake configure aborts at `test/cmake/CodeCoverage.cmake:130` with `lcov not found!`, which has no Amazon Linux 2 package, so `-DDISABLE_TESTS=ON` is needed there. The C `ctest` suite was therefore not run.

Environment: GraalVM CE 21.0.2, CMake 3.28.3, .NET SDK 8.0.425, Corretto 8 for the Maven builds, Amazon Linux 2.

knoxy5467 (approved):

PR #526, opened 2026-06-30
Fix deserialization of optional union types (anyOf null) (fixes #218)

### LINKED ISSUE #218: Issue deserializing optional union types
We are having issues deserializing the data with the following schema:
`{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "name": {
      "type": [
        "string",
        "null"
      ]
    }
  },
  "additionalProperties": false
}`
with this data:
`{
    "name": "Cristina Hermann"
}`
I believe it is because of this check here: https://github.com/awslabs/aws-glue-schema-registry/blob/master/jsonschema-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/jsonschema/JsonSchemaToConnectSchemaConverter.java#L96
I think this check should be also including ANY_OF criterion.

shakhihali:
## Description

Fixes deserialization of JSON Schema optional union types declared via a
`type` array, e.g.:

```json
{ "type": "object", "properties": { "name": { "type": ["string", "null"] } } }
```

Fixes #218.

## Root cause

`JsonSchemaToConnectSchemaConverter#toConnectSchema` detects an "optional union"
(a single real type plus `null`) only when the everit `CombinedSchema` uses the
`ONE_CRITERION` (`oneOf`):

```java
boolean isOptionalUnion =
    CombinedSchema.ONE_CRITERION.equals(criterion) && subSchemas.size() == 2 && hasNullSchema;
```

However, a `"type": ["string", "null"]` array is parsed by everit as a
`CombinedSchema` using the `ANY_CRITERION` (`anyOf`), not `oneOf`. It therefore
fails this check and falls through to `buildNonOptionalUnionSchema`, which calls
`builder.optional()`. Control then returns to `populateConnectProperties`, which
calls `builder.required()` on the same builder (struct fields are converted with
`required = true`), and Kafka Connect's `SchemaBuilder` rejects it:

```
org.apache.kafka.connect.errors.SchemaBuilderException:
    Invalid SchemaBuilder call: optional has already been set.
```

## Fix

Treat the optional-union shape (two subschemas, one of which is `NullSchema`) as
optional regardless of whether everit modeled it as `oneOf` or `anyOf`:

```java
boolean isOptionalUnion =
    (CombinedSchema.ONE_CRITERION.equals(criterion) || CombinedSchema.ANY_CRITERION.equals(criterion))
        && subSchemas.size() == 2 && hasNullSchema;
```

This matches the maintainer-independent diagnosis in the original issue (the
check "should also include the ANY_OF criterion").

## Testing

Added two regression tests to `ToConnectTest` using the exact schema and data
from the issue:

- `testToConnect_optionalUnionTypeArray_doesNotThrow` — `{"name": "Cristina Hermann"}`
  converts without throwing; the `name` field is an optional `STRING`.
- `testToConnect_optionalUnionTypeArray_withNullValue_doesNotThrow` —
  `{"name": null}` is accepted and validates.

Both tests fail on `master` with the exact `SchemaBuilderException` from the
issue and pass with this change.

- `mvn -pl jsonschema-kafkaconnect-converter test` — 331 tests, 0 failures
  (includes the existing `FromConnectTest` / `JsonSchemaConverterTest` round-trip
  suites that exercise the `oneOf` union path; no regressions).
- `mvn clean verify` — BUILD SUCCESS across all 13 modules.

### DIFF
--- a/jsonschema-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/jsonschema/JsonSchemaToConnectSchemaConverter.java
+++ b/jsonschema-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/jsonschema/JsonSchemaToConnectSchemaConverter.java
@@ -33,7 +33,6 @@ import org.everit.json.schema.ReferenceSchema;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.Collectors;
 
 /**
@@ -92,16 +91,19 @@ public class JsonSchemaToConnectSchemaConverter {
             boolean hasNullSchema = subSchemas.stream()
                     .anyMatch(schema -> schema instanceof NullSchema);
 
-            // An optional union is a two-subschema combination where one subschema is
-            // NullSchema, e.g. {"type": ["string", "null"]}. Depending on how the JSON
-            // Schema is authored, everit represents this either as oneOf (ONE_CRITERION)
-            // or anyOf (ANY_CRITERION) - notably, a "type" array is parsed as anyOf. Both
-            // shapes describe the same optional, single-type field, so detect either one.
+            // A nullable union is a oneOf/anyOf that includes a NullSchema, e.g.
+            // {"type": ["string", "null"]} or {"type": ["string", "integer", "null"]}.
+            // (A "type" array is parsed by everit as anyOf; an explicit oneOf uses
+            // ONE_CRITERION.) In all these cases the null branch means the field is
+            // optional, and the field should carry the union of the remaining, non-null
+            // types. Build that optional schema and return it directly - it must not fall
+            // through to populateConnectProperties, which would call required() on an
+            // already-optional builder and throw "optional has already been set".
             // See https://github.com/awslabs/aws-glue-schema-registry/issues/218
-            boolean isOptionalUnion =
-                    (CombinedSchema.ONE_CRITERION.equals(criterion) || CombinedSchema.ANY_CRITERION.equals(criterion))
-                            && subSchemas.size() == 2 && hasNullSchema;
-            if (isOptionalUnion) {
+            boolean isNullableUnion = hasNullSchema
+                    && (CombinedSchema.ONE_CRITERION.equals(criterion)
+                            || CombinedSchema.ANY_CRITERION.equals(criterion));
+            if (isNullableUnion) {
                 return buildOptionalUnionSchema(subSchemas);
             }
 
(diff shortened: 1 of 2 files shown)


knoxy5467:
Requesting changes. The diagnosis is right and the `anyOf` handling is genuinely needed, but this only fixes the two-type case and leaves the same bug in place for anything wider.

`["string", "null"]` parses to a `CombinedSchema` with two subschemas, so it passes the new `subSchemas.size() == 2` check and gets treated as optional. But a nullable union with more than one real type is just as valid — `["string", "integer", "null"]` is a normal schema — and that comes through as three subschemas. It fails the `== 2` check, falls into `buildNonOptionalUnionSchema` (which calls `optional()` because of the null branch), and then the field's `required()` is applied on the same builder, so you get the exact `SchemaBuilderException: optional has already been set` that #218 is about. Same failure, slightly different schema.

So as written this closes the one schema in the report but not the underlying problem — the next person with a three-way nullable union hits the identical error and files the same issue. I'd like the fix to handle the general "one or more real types plus null" shape, not just the two-subschema case: pull the null out, mark the field optional, and build the union from however many non-null subschemas remain. A test with `["string", "integer", "null"]` would lock it in.

shakhihali:
Good catch — you're right, and I've generalized the fix.

Confirmed the gap first: I added a `["string", "integer", "null"]` test and it threw the exact `SchemaBuilderException: optional has already been set` you described. The old code only returned early for `subSchemas.size() == 2`; a three-way nullable union fell through to `buildNonOptionalUnionSchema` (which sets `optional()`) and then had `required()` applied on the same builder.

The fix now treats **any** nullable union as optional rather than the two-subschema special case:

- `isNullableUnion = hasNullSchema && (oneOf || anyOf)` — no size check.
- `buildOptionalUnionSchema` pulls out the null and builds from the remaining non-null subschemas: one real type → that type made optional (e.g. `["string","null"]` → optional STRING); two or more → the oneOf-style union struct over the real types, marked optional and returned directly so it doesn't hit `required()`.

Tests added for `["string", "integer", "null"]` (schema construction no longer throws; the field is an optional 2-field union struct) and for the explicit-null value. Existing `["string","null"]` tests still pass. `mvn -pl jsonschema-kafkaconnect-converter test` → 333 tests, 0 failures; `mvn clean verify` green across all modules.

One thing I want to flag honestly, since I ran into it while testing: deserializing a *non-null* value into a multi-type union (e.g. `{"id": 42}`) currently fails, but that's a **pre-existing, separate** issue in the union value matcher in `StructTypeConverter` — it reproduces on a plain non-nullable `["string","integer"]` union on master too (integer values don't match the `FLOAT64` field's node-type list, and non-matching union fields are built as required). It's independent of the `SchemaBuilderException` this PR/issue is about, so I've kept it out of scope rather than expand this change silently. Happy to file/fix it separately if you'd like — let me know and I'll open a follow-up.

jvdadda:
Hi @shakhihali, good diagnosis, and generalising past the two-type case after @knoxy5467's review was the right call. We ported your fix into our maintained fork of the Java library in https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/116; it is available in the latest published release, currently 4.0.0. Same shape as your final version: any `oneOf` / `anyOf` / `"type": [...]` containing `null` is nullable whatever its arity (one real type becomes that type, optional; several become the optional union struct), with the three-way `["string", "integer", "null"]` case among the tests.

It is on Maven Central as `com.mobsuccess:jsonschema-kafkaconnect-converter`; Connect workers take the `-plugin.zip` attached to each release instead. Your #527 (`const`) is on our list too. The whole upstream/fork correspondence is kept in https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/129.

PR #527, opened 2026-06-30
Support JSON Schema const types in Kafka Connect converter

shakhihali:
## Description

Adds support for JSON Schema `const` types to the Kafka Connect JSON Schema
converter. Reported via customer feedback: the Kafka Connect translation failed
on schemas using `const`.

## Problem

On `master`, any `const` schema fails immediately with:

```
org.apache.kafka.connect.errors.DataException:
    Unsupported schema type org.everit.json.schema.ConstSchema
```

## Approach (reworked per review feedback)

A `const` restricts a value to a single permitted value, which per draft-06+ may
be **any** JSON value — scalar, object, or array. everit exposes it as a
`ConstSchema` whose `getPermittedValue()` returns the value with its natural Java
type (`String`, `Integer`, `Boolean`, `Map`, `List`, ...).

The converter now handles `ConstSchema` in `JsonSchemaToConnectSchemaConverter`
by **recursively inferring** a Connect schema from the shape of the permitted
value:

- scalar → matching Connect type (`STRING`, `INT64`, `FLOAT64`, `BOOLEAN`)
- object (`Map`) → `STRUCT`, recursing per field (fields sorted for deterministic order)
- array (`Collection`) → `ARRAY` with a homogeneous element type, recursing on the element

Because value deserialization is driven by the Connect schema type, routing
through the normal type machinery means both schema translation **and** value
deserialization work for objects and arrays — not just scalars.

Ambiguous cases that cannot map to a Connect schema fail fast with a clear,
specific message instead of a misleading downstream error:

- empty `const` array (no element type to infer)
- heterogeneous `const` array (Connect `ARRAY` requires a single element type)
- `null` const value / `null` object-field value (no inferable type)

### Scope / honest limitation

This preserves the **type** of the const value so data flows through correctly.
It does **not** enforce the const **constraint** (that the value must equal the
permitted value), and the constraint is not round-tripped back to `const` — Connect's
schema model has no equality constraint. This is consistent with how the converter
already treats other JSON Schema validation keywords (`pattern`, `minimum`,
`format`, ...): the type is kept, the constraint is dropped.

## What changed since the first version of this PR

The initial version mapped `const` in `TypeConverterFactory` using only the
scalar Java type, which (correctly flagged in review) mis-typed object/array
consts as `STRING` and produced a confusing downstream
`Invalid null value for required STRING field` — strictly worse than master's
clear "unsupported" error. That approach has been fully reverted; `const` is now
handled at the schema-construction layer with recursive inference, and the
object/array cases are covered by tests.

## Testing

Added tests in `ToConnectTest` covering each shape and the fail-fast cases:
string, integer, number, boolean, object→STRUCT, array→ARRAY, nested
object+array, object-value deserialization into a `Struct`, and expected
`DataException`s for empty and heterogeneous arrays.

- `mvn -pl jsonschema-kafkaconnect-converter test` — 339 tests, 0 failures.
- `mvn clean verify` — BUILD SUCCESS across all 13 modules.

### DIFF
--- a/jsonschema-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/jsonschema/JsonSchemaToConnectSchemaConverter.java
+++ b/jsonschema-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/jsonschema/JsonSchemaToConnectSchemaConverter.java
@@ -27,13 +27,16 @@ import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.errors.DataException;
 import org.everit.json.schema.CombinedSchema;
+import org.everit.json.schema.ConstSchema;
 import org.everit.json.schema.NullSchema;
 import org.everit.json.schema.ReferenceSchema;
 
+import java.math.BigInteger;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 
 /**
@@ -84,6 +87,13 @@ public class JsonSchemaToConnectSchemaConverter {
 
         if (typeConverter != null) {
             builder = typeConverter.toConnectSchema(jsonSchema, jsonSchemaDataConfig);
+        } else if (jsonSchema instanceof ConstSchema) {
+            // JSON Schema "const" (draft-06+) restricts a value to a single permitted
+            // value, which may be any JSON value - scalar, object, or array. everit exposes
+            // that value via getPermittedValue(). We infer a Connect schema from the value's
+            // shape so both schema translation and value deserialization work through the
+            // normal Connect type machinery. See issue: const types unsupported.
+            builder = buildConstSchema((ConstSchema) jsonSchema);
         } else if (jsonSchema instanceof CombinedSchema) {
             CombinedSchema combinedSchema = (CombinedSchema) jsonSchema;
             Collection<org.everit.json.schema.Schema> subSchemas = combinedSchema.getSubschemas();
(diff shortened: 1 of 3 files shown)


knoxy5467:
I'd hold off on merging this one. My worry is that it ends up worse than what's on master today.

Right now if you hand the converter a `const`, it throws `Unsupported schema type ...ConstSchema` right away. Users immediately know the converter can't handle it. With this change, only string/number/boolean consts actually work. The problem is `const` isn't limited to scalars; per draft-06 it can be any JSON value, so things like `{"const": {"code": "US"}}` or `{"const": ["a","b"]}` are perfectly valid. everit hands those back as a `JSONObject`/`JSONArray`, and `getForConstValue` only checks for `Boolean` and `Number` before falling through to `get(STRING)`. So an object const gets typed as a STRING, and then on the value side `StringTypeConverter.toConnect` calls `value.textValue()`, which is `null` for an object node, because the field is required, you end up with `Invalid null value for required STRING field`. Instead of a clear "can't do this," you get a wrong type and a confusing failure further down (or silently dropped data). 

I'm also not sure `const` belongs in the type factory at all. It's a validation constraint, same as `pattern`, `minimum`/`maximum`, `format`. The converter already drops all of those, because Connect's schema model has no way to say "this value must equal X." This PR doesn't preserve the constraint either: `{"const": 1}` becomes an `INT64`, and coming back out it's just `{"type":"integer"}`, with nothing recording that it was ever pinned to 1. So we're adding a type-factory branch for something that isn't really a type and can't round-trip anyway.

And for the object/array case there isn't a clean way to do it. `getPermittedValue()` gives you one literal value, so to produce a STRUCT you'd have to infer field types.

const field always carries the same value, and that value already flows through fine as normal record data. If we do want const support down the line, we need to address the complex type validation for all objects validated by const.

shakhihali:
Thanks — this was the right call, and I've reworked the PR accordingly.

You were correct that the original approach was a net regression: I reproduced it, and an object-valued const like `{"const": {"code": "US"}}` got typed as `STRING` and then failed downstream with `Invalid value: null used for required field ... STRING` — strictly worse than the clear "unsupported schema type" you get on master today. I've fully reverted the `TypeConverterFactory` change.

The new approach handles `const` at the schema-construction layer (`JsonSchemaToConnectSchemaConverter`) by recursively inferring the Connect schema from `getPermittedValue()`:

- object (`Map`) → `STRUCT`, recursing per field
- array (`Collection`) → `ARRAY` with a homogeneous element type
- scalars → their Connect types

Since the value path is driven by the Connect schema type, this makes object and array consts actually deserialize, not just scalars. The genuinely ambiguous cases you'd otherwise hit silently now fail fast with specific messages: empty array, heterogeneous array, and null-typed values.

On your second point — that `const` is a validation constraint, not a type — you're right, and I'm not claiming otherwise. This preserves the const value's **type** so the data flows through; it does **not** enforce or round-trip the constraint, exactly as the converter already drops `pattern`/`minimum`/`format`. I've said this explicitly in the PR description and a code comment so it's not mistaken for full const semantics.

Tests added for each shape (scalar/object/array/nested), object-value deserialization into a `Struct`, and the two fail-fast array cases. `mvn -pl jsonschema-kafkaconnect-converter test` → 339 tests, 0 failures; `mvn clean verify` green across all modules.

If you'd rather this not go in at all until there's a plan for constraint enforcement, I'm fine with that too — but this at least turns the object/array case from "confusing failure" into "works," which was the core of your concern.

jvdadda:
Hi @shakhihali, the reworked version of this PR is the right one, and it is now in our maintained fork of the Java library as https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/171, with your tests transcribed as they are: `const` is handled at schema construction in `JsonSchemaToConnectSchemaConverter`, the Connect schema inferred recursively from the permitted value, so an object or array const converts instead of being mistyped as a string, which was @knoxy5467's concern with the first version. It is available in the latest published release, currently 4.1.0, on Maven Central as `com.mobsuccess:jsonschema-kafkaconnect-converter`; Connect workers take the `-plugin.zip` attached to the release.

Issue #275, opened 2023-07-07
Schema converters: add caching of the schema object to improve performance in 10x times


gorbenkoya:
**Good day, I have found major issue related to performance of the jsonschema-kafkaconnect-converter and probably other converters** 

In the company where I work we use it as part of our KC S3 Parquet connector setup to write millions of files into S3 in parquet format with the schema located in the Glue registry. 
After initial deployment of KC Parquet connector we have noticed it's incredibly slow, writing only a few thousand records per minute, adding connector workers and KC nodes didn't help a lot. 
After several back and forth I have find out they cause of such slowness. The problem lives in the schema caching which is currently done, but on the very basic level, it's cached as String. It means that for every incoming record schema should be constructed from String to the Schema object by the following lines of code:
 
`   JSONObject jsonSchemaObject = new JSONObject(jsonSchemaString);
            jsonSchema = SchemaLoader.builder()
                   .schemaJson(jsonSchemaObject)
                    .build()
                   .load()
                    .build();`  

This schema construction operation is incredibly heavy, so it slows down converter so much. 
As the solution I have added caching for the top level schema object. 

**As the result overall performance of the connector has been increased in 10-20 times!** 
  
Please see proposed fix in the attachment. 
Please review other converters for the similar issue and fix it.    

[JsonSchemaConverter_PerformanceX10.diff.txt](https://github.com/awslabs/aws-glue-schema-registry/files/11986670/JsonSchemaConverter_PerformanceX10.diff.txt)

blacktooth:
Thanks for raising this isse! We will analyze the findings and see if it can be incorporated into the code base. If we decide to proceed, would you be willing to submit a PR?

djchapm:
We're seeing slow cache performance as well just using the aws sdk, same issue?  

Any updates?

PR #528, opened 2026-07-03
Cache parsed Avro Schema to fix performance issue (fixes #478)

### LINKED ISSUE #478: Glue schema Avro serdes - performance issue (Kafka Connect & Flink)
We are reporting serious performance degradation with usage of Glue Schema Registry Avro serializer / deserializer.
Observed for: Flink and Kafka Connect implementations.

Points in code where we have detected issues:
Kafka Connect:
* [https://github.com/awslabs/aws-glue-schema-registry/blob/master/avro-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/AWSKafkaAvroConverter.java#L165 ](https://github.com/awslabs/aws-glue-schema-registry/blob/master/avro-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/AWSKafkaAvroConverter.java#L165)

Flink:
* [https://github.com/awslabs/aws-glue-schema-registry/blob/master/avro-flink-serde/src/main/java/com/amazonaws/services/schemaregistry/flink/avro/GlueSchemaRegistryInputStreamDeserializer.java#L74 ](https://github.com/awslabs/aws-glue-schema-registry/blob/master/avro-flink-serde/src/main/java/com/amazonaws/services/schemaregistry/flink/avro/GlueSchemaRegistryInputStreamDeserializer.java#L74)

Doing the schema parsing from String to the object: org.apache.avro.Schema for every processing event is very expensive operation, and it is root cause of performance problems.
We have measured and proved major performance improvement (3,5 - 4 times faster processing) with exclusion of schema parsing from String.
Apart of gain the processing speed the CPU allocation is significantly lower (and saving the costs).

awsmasudur:
## Summary
Adds a ConcurrentHashMap cache for parsed `org.apache.avro.Schema` objects in the Flink and Kafka Connect Avro deserializers. This avoids calling `Schema.Parser().parse()` on every event, which was identified as a 3.5-4x performance bottleneck in #478.

## Changes
- **avro-flink-serde**: `GlueSchemaRegistryInputStreamDeserializer` now caches parsed schemas keyed by schema definition string
- **avro-kafkaconnect-converter**: `AWSKafkaAvroConverter` now caches parsed schemas keyed by schema definition string

## Testing
- All existing unit tests pass (22 tests in avro-kafkaconnect-converter, 20 tests in avro-flink-serde)
- No API changes, drop-in behavioral improvement

Fixes #478

### DIFF
--- a/avro-flink-serde/src/main/java/com/amazonaws/services/schemaregistry/flink/avro/GlueSchemaRegistryInputStreamDeserializer.java
+++ b/avro-flink-serde/src/main/java/com/amazonaws/services/schemaregistry/flink/avro/GlueSchemaRegistryInputStreamDeserializer.java
@@ -27,6 +27,7 @@ import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 
 /**
@@ -35,6 +36,7 @@ import java.util.Map;
  */
 public class GlueSchemaRegistryInputStreamDeserializer {
     private final GlueSchemaRegistryDeserializationFacade glueSchemaRegistryDeserializationFacade;
+    private final ConcurrentHashMap<String, Schema> parsedSchemaCache = new ConcurrentHashMap<>();
 
     /**
      * Constructor accepts configuration map for AWS Deserializer
@@ -69,13 +71,14 @@ public class GlueSchemaRegistryInputStreamDeserializer {
         byte[] deserializedBytes = glueSchemaRegistryDeserializationFacade.getActualData(inputBytes);
         mutableByteArrayInputStream.setBuffer(deserializedBytes);
 
-        Schema schema;
-        try {
-            schema = (new Schema.Parser()).parse(schemaDefinition);
-        } catch (SchemaParseException e) {
-            String message = "Error occurred while parsing schema, see inner exception for details.";
-            throw new AWSSchemaRegistryException(message, e);
-        }
+        Schema schema = parsedSchemaCache.computeIfAbsent(schemaDefinition, schemaDef -> {
+            try {
+                return (new Schema.Parser()).parse(schemaDef);
+            } catch (SchemaParseException e) {
+                String message = "Error occurred while parsing schema, see inner exception for details.";
+                throw new AWSSchemaRegistryException(message, e);
+            }
+        });
 
         return schema;
     }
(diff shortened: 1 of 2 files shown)


knoxy5467:
How does this interact with the existing caching behavior?

avro-flink-serde/src/main/java/com/amazonaws/services/schemaregistry/flink/avro/GlueSchemaRegistryInputStreamDeserializer.java
```diff
@@ -69,13 +71,14 @@ public Schema getSchemaAndDeserializedStream(InputStream in) throws IOException
...
         mutableByteArrayInputStream.setBuffer(deserializedBytes);
 
-        Schema schema;
-        try {
-            schema = (new Schema.Parser()).parse(schemaDefinition);
-        } catch (SchemaParseException e) {
-            String message = "Error occurred while parsing schema, see inner exception for details.";
-            throw new AWSSchemaRegistryException(message, e);
-        }
+        Schema schema = parsedSchemaCache.computeIfAbsent(schemaDefinition, schemaDef -> {
```
michalklempa, line 74: This uses the whole schemaDefinition as a key into HashMap. That might be next performance bottleneck. But.

It is much easier to see if this happens or not, when the implementation does not use the schemaString as a [key into hashmap](https://github.com/confluentinc/schema-registry/blob/v7.7.5/client/src/main/java/io/confluent/kafka/schemaregistry/client/CachedSchemaRegistryClient.java#L488)

Unfortunately in 8.x branch on Confluent schema-registry client, it is a bit more complicated and hidden in a [2 level caching, and using the content as the key](https://github.com/confluentinc/schema-registry/blob/v8.3.1-40/client/src/main/java/io/confluent/kafka/schemaregistry/client/CachedSchemaRegistryClient.java#L613
).

But it relies on the first cache by id to return the same string instance each time so it literally relies on JDK implementation caching the String instance [hash code](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/String.java#L2788)
The schemaDefinition is already cached in Facade first-level UUID -> Schema cache.

Now this implementation seems to do something similar, first-level cache is in Facade storing the String instance and here we rely on hashCode of the String to be cached inside String instance therefore using the large schema String as a key might be OK.

There are no tests proving/disproving neither performance tests.
Have we checked this at least using debugging?

michalklempa:
> How does this interact with the existing caching behavior?

I do [believe](https://github.com/awslabs/aws-glue-schema-registry/pull/528/changes#r3585621018) that it actually silently relies on existing caching mechanism. More tests might be needed.

jvdadda:
Hi @awsmasudur, thanks for this one: it pinned down exactly the right hot spot for #478 and #275. While it waits here, we carried it into our maintained fork of the Java library (Kotlin sources, same packages and configuration keys) in https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/91. It is available in the latest published release, currently 4.0.0, on Maven Central as `com.mobsuccess:schema-registry-kafkaconnect-converter` and `com.mobsuccess:schema-registry-flink-serde`.

On @knoxy5467's question, as we understood it: the registry lookup and its existing cache are untouched (`getSchemaDefinition` is still called for every record, and a test asserts that), and only the `Schema.Parser().parse()` step is memoised, in a bounded Guava cache of 100 entries, the same bound the datum reader/writer caches already use, rather than an unbounded map. Measured in isolation the parse costs about 200x the cache hit, so the gain grows with schema size.

PR #529, opened 2026-07-03
Add support for nullable fields in JSON Schema generation (fixes #73)

### LINKED ISSUE #73: Support for nullable fields in JsonSchema generation from POJO's
When sending POJO's using the JSON data format, the schema that gets generated do not allow for nullable fields. There is also no way to pass this intent as part of the configuration.

In particular, I'm referring to the following:
`com.amazonaws.services.schemaregistry.serializers.json.JsonSerializer`  

serializes using the default configuration

```this.jsonSchemaGenerator = new JsonSchemaGenerator(this.objectMapper);```

However, the `mbknor-jackson-jsonSchema` library used has a configuration that, when generating a schema allows for nullable files , see the `Nullable types` section [here](https://github.com/mbknor/mbknor-jackson-jsonSchema)
```
JsonSchemaConfig config = JsonSchemaConfig.nullableJsonSchemaDraft4();
JsonSchemaGenerator generator = new JsonSchemaGenerator(objectMapper, config);
```

This library needs to be extended with a configuration to allow for the generation of a schema that supports nullable types as described above.

without nullable fileds
```
"properties": {
        "addressId": {
          "type": "string"
        },
```

vs

with nullable fields
```
"properties": {
        "addressId": {
	      "oneOf": [
		      {
			  "type": "null",
			  "title": "Not included"
		      },
		      {
			    "type": "string"
		      }
	      ]
        },
```

### PR DESCRIPTION
**Pull request by awsmasudur**
## Summary
Adds a new configuration property `jsonSchemaNullableEnabled` that enables nullable field support in JSON Schema generation from POJOs. When enabled, the `mbknor-jackson-jsonSchema` library uses `JsonSchemaConfig.nullableJsonSchemaDraft4()` which generates schemas with `oneOf[null, type]` for nullable fields.

## Changes
- **AWSSchemaRegistryConstants.java**: Added `JSON_SCHEMA_NULLABLE_ENABLED` constant
- **GlueSchemaRegistryConfiguration.java**: Added `jsonSchemaNullableEnabled` boolean field (default `false`) with validation
- **JsonSerializer.java**: Conditionally uses nullable JSON schema config when enabled

## Usage
`java
Map<String, Object> configs = new HashMap<>();
configs.put(region, us-east-1);
configs.put(jsonSchemaNullableEnabled, true);
// Schema generation will now produce oneOf[null, type] for nullable fields
``n
## Testing
- All existing unit tests pass (1332 tests in serializer-deserializer, 118 tests in common)
- Default behavior is unchanged (disabled by default)

Fixes #73

### DIFF
--- a/common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
+++ b/common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
@@ -67,6 +67,7 @@ public class GlueSchemaRegistryConfiguration {
 
     private List<SerializationFeature> jacksonSerializationFeatures;
     private List<DeserializationFeature> jacksonDeserializationFeatures;
+    private boolean jsonSchemaNullableEnabled = false;
 
     public GlueSchemaRegistryConfiguration(String region) {
         Map<String, Object> config = new HashMap<>();
@@ -99,6 +100,7 @@ public class GlueSchemaRegistryConfiguration {
         validateAndSetSchemaAutoRegistrationSetting(configs);
         validateAndSetJacksonSerializationFeatures(configs);
         validateAndSetJacksonDeserializationFeatures(configs);
+        validateAndSetJsonSchemaNullableEnabled(configs);
         validateAndSetTags(configs);
         validateAndSetMetadata(configs);
         validateAndSetUserAgent(configs);
@@ -320,6 +322,13 @@ public class GlueSchemaRegistryConfiguration {
             } else {
                 throw new AWSSchemaRegistryException("Jackson Deserialization features should be a list");
             }
+        }
+    }
+
+    private void validateAndSetJsonSchemaNullableEnabled(Map<String, ?> configs) {
+        if (isPresent(configs, AWSSchemaRegistryConstants.JSON_SCHEMA_NULLABLE_ENABLED)) {
+            this.jsonSchemaNullableEnabled = Boolean.parseBoolean(
+                    configs.get(AWSSchemaRegistryConstants.JSON_SCHEMA_NULLABLE_ENABLED).toString());
         }
     }
 
--- a/common/src/main/java/com/amazonaws/services/schemaregistry/utils/AWSSchemaRegistryConstants.java
+++ b/common/src/main/java/com/amazonaws/services/schemaregistry/utils/AWSSchemaRegistryConstants.java
@@ -167,6 +167,13 @@ public final class AWSSchemaRegistryConstants {
     public static final String JACKSON_DESERIALIZATION_FEATURES = "jacksonDeserializationFeatures";
 
     /**
+     * Configuration to enable nullable fields in JSON Schema generation.
+     * When enabled, uses JsonSchemaConfig.nullableJsonSchemaDraft4() which generates
+     * schemas with oneOf[null, type] for nullable fields.
+     */
+    public static final String JSON_SCHEMA_NULLABLE_ENABLED = "jsonSchemaNullableEnabled";
+
+    /**
      * Name of the UserAgent application using the serializer/de-serializer library.
      * Ex: KPL, Kafka, KafkaConnect etc.
      */
--- a/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/json/JsonSerializer.java
+++ b/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/serializers/json/JsonSerializer.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.kjetland.jackson.jsonSchema.JsonSchemaConfig;
 import com.kjetland.jackson.jsonSchema.JsonSchemaGenerator;
 import lombok.Builder;
 import lombok.Getter;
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by knoxy5467** [comment]
Can we have an example use case or set of schemas where this would be used? I want to confirm that this is not a breaking change.

**Comment by knoxy5467** [comment]
Additionally why does the JsonProperty Annotations that already exists https://github.com/awsmasudur/aws-glue-schema-registry#dealing-with-specific-record-java-pojo-for-json . Not address this already (setting required = false)?

**Comment by knoxy5467** [comment]
After looking even further. While this would allow support for fields marked non-required to be a null union type. I would prefer that if we are going to add addtional options. we add the capability to fully pass in the config object for the generator instead of having it be one or the other.

Technically it is not needed because users can still use the @JsonSchemaInject annotation.

**Comment by jvdadda** [comment]
Thanks for this one as well: it is the answer to #73, open since 2022. We ported it into our fork in https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/117, and it is available in the latest published release, currently 4.0.0 (`com.mobsuccess:schema-registry-serde` on Maven Central).

Same property name (`jsonSchemaNullableEnabled`), same default of `false`, same `nullableJsonSchemaDraft4()` switch, and we kept your reasoning for opt-in: turning it on changes the generated schema text, hence registers a new schema version, so it should be something a producer chooses rather than something an upgrade does to it. We did keep it as a single flag rather than exposing the whole generator config as @knoxy5467 suggested; that remains a reasonable follow-up. The property is documented in https://github.com/mobsuccess-devops/aws-glue-schema-registry/blob/master/docs/configuration.md.

**Comment by awsmasudur** [comment]
Thanks for the thorough review @knoxy5467 — I've pushed an update that addresses all three points.

**1. Is this a breaking change? (example use case + schemas)**

No. The feature is disabled by default, so existing schema generation is byte-for-byte identical. I added a regression test asserting the default output is unchanged (`testPojo_getSchemaDefinition_nullableDisabledByDefault_unchanged`), so this is now guaranteed in CI rather than just by inspection.

Before/after for a non-required `String` field:

```jsonc
// default (jsonSchemaNullableEnabled = false)
"addressId": { "type": "string" }

// jsonSchemaNullableEnabled = true
"addressId": {
  "oneOf": [
    { "type": "null", "title": "Not included" },
    { "type": "string" }
  ]
}
```

Enabling it does change the generated schema text, which registers a new schema version — that is exactly why it's opt-in per producer rather than something an upgrade does automatically.

**2. Why doesn't `@JsonProperty(required = false)` already handle this?**

`required` and nullability are orthogonal in the generator. `required = false` only removes a field from the schema's `required` array (controls whether the key must be *present*); it does not change the field's type, so the value still cannot be `null` — `{"type":"string"}` rejects `null`. Making the value nullable is a separate toggle (`useOneOfForNullables`, which is what `nullableJsonSchemaDraft4()` flips). Per the library's own docs, `@JsonProperty(required = true)` is actually the way to *suppress* nullability on a specific field once the feature is enabled — which this PR supports.

**3. "Prefer passing in the full config object instead of one-or-the-other" + `@JsonSchemaInject` exists**

Agreed, and done. In addition to the simple `jsonSchemaNullableEnabled` flag for the common case, you can now pass a fully-customized `com.kjetland.jackson.jsonSchema.JsonSchemaConfig` via the `jsonSchemaConfig` property. When present it takes precedence over the flag, so nullable support can be combined with any other generator option rather than being an either/or preset:

```java
configs.put(AWSSchemaRegistryConstants.JSON_SCHEMA_CONFIG, JsonSchemaConfig.nullableJsonSchemaDraft4());
```

On `@JsonSchemaInject`: it's a valid manual escape hatch, but for nullability it forces hand-written `oneOf[null, …]` fragments on every nullable field of every POJO and doesn't cover `Optional<T>`, so a generator-level option still earns its place. The config-object path above gives the full flexibility you asked for without that per-field boilerplate.

**Also added:** unit tests (config parsing, default-unchanged, nullable `oneOf`, config-object precedence, wrong-type rejection) and README documentation with the before/after example.

**One design note worth flagging:** the strongly-typed `jsonSchemaConfig` property means the `common` module now takes a compile dependency on `mbknor-jackson-jsonschema` (it previously had none). That's the cost of type safety and fail-fast validation. If you'd rather keep `common` free of that dependency, the fallback is to hold the field as `Object` and cast it in the serializer — happy to switch if you prefer.

Full `mvn verify` passes on both `common` and `serializer-deserializer` (checkstyle clean, coverage gates met).

awsmasudur:
Heads up on the red **Build CSharp / build-musl (x64)** check — it is a pre-existing upstream failure, not something this PR introduces.

The failing step is *Build C components with musl*, which errors on a clang-tidy rule (warnings treated as errors) in the native C layer:

```
multilang-schema-registry/c/src/glue_schema_registry_deserializer.c:150:19:
error: implicit conversion 'char' -> 'bool' [readability-implicit-bool-conversion,-warnings-as-errors]
  150 |     bool result = can_decode(thread, array, p_err);
1 warning treated as error
```

A couple of points:

- This PR is Java-only (plus README and pom) — it touches **no** C/C++ files, so it cannot cause a C compilation/lint failure. The *Build Java components with musl* step in the same job passes.
- The same `Build CSharp` workflow is already failing on `master` independently of this PR (e.g. the latest master run for the #539 CVE merge is red on this workflow). The failure was inherited when I merged the latest `master` into this branch to clear the earlier merge conflicts.

So the fix belongs in a separate, C-focused change on `master` (making the `char` → `bool` conversion explicit at line 150), rather than in this JSON-schema PR. Happy to open that separately if it would help, but I have kept this PR scoped to the nullable-schema feature.

The Java/`common`/`serializer-deserializer` builds pass locally under `mvn verify` (checkstyle clean, coverage gates met).

knoxy5467:
We don't have any plans to merge this at the moment because it breaks the contract of what Kafka Connect and JSON schemas currently support. Internal discussions on how to move forward are ongoing.

knoxy5467:
> No. The feature is disabled by default, so existing schema generation is byte-for-byte identical. I added a regression test asserting the default output is unchanged (`testPojo_getSchemaDefinition_nullableDisabledByDefault_unchanged`), so this is now guaranteed in CI rather than just by inspection.

We don't know if we want to support and additional feature along these lines yet. Json schema supports more than just nullable fields and allows for union of any types and multiple types. Kafka connect schema has no way to support this

PR #530, opened 2026-07-03
Bump kotlinx-serialization-core-jvm from 1.4.0 to 1.7.3 (fixes #313)

### LINKED ISSUE #313: Dependency conflict prevents usage of this library if a project has newer kotlinx-serialization

```
}
    version { strictly("1.6.1") }
}
```

* Gradle Version: 8.3

awsmasudur:
## Summary
Bumps `kotlinx-serialization-core-jvm` from 1.4.0 to 1.7.3 to resolve dependency conflicts for projects using newer Kotlin (1.9.x+) with kotlinx.serialization 1.6+.

## Problem
The pinned 1.4.0 version caused conflicts for downstream users whose projects depend on newer kotlinx.serialization versions, resulting in the error: *Your current kotlinx.serialization core version is too low*.

## Changes
- **serializer-deserializer/pom.xml**: Updated `kotlinx-serialization-core-jvm` version from `1.4.0` to `1.7.3`

## Compatibility
- Version 1.7.3 is compatible with the project's Kotlin 1.9.25 and Wire 5.2.0
- kotlinx.serialization maintains backward compatibility
- This dependency is not used directly in Java source code; it's a transitive requirement for Wire's protobuf support

## Testing
- All 1332 existing unit tests pass with 0 failures

Fixes #313

### DIFF
--- a/serializer-deserializer/pom.xml
+++ b/serializer-deserializer/pom.xml
@@ -159,7 +159,7 @@
 	<dependency>
 	    <groupId>org.jetbrains.kotlinx</groupId>
 	    <artifactId>kotlinx-serialization-core-jvm</artifactId>
-	    <version>1.4.0</version>
+	    <version>1.7.3</version>
 	</dependency>
 	<dependency>
             <groupId>com.squareup.wire</groupId>


knoxy5467:
Does 1.7.3 have the potential to break any consumers of the existing library or is it completely backwards compatible?

awsmasudur:
Thanks for raising this. I checked it rather than just assuming, and it's backward compatible for consumers of the library.

A couple of reasons:

- `kotlinx-serialization-core-jvm` isn't referenced anywhere in the project's source. It's declared only in `serializer-deserializer/pom.xml` as a transitive/runtime requirement for Square Wire's protobuf handling, so it isn't part of any public API the library exposes. Bumping it doesn't change any signatures or behavior consumers depend on.
- kotlinx.serialization guarantees binary backward compatibility for its stable API since 1.0.0 ([compatibility policy](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/compatibility.md)). Going from 1.4.0 to 1.7.3 stays within that guarantee, and the only surface without a compatibility guarantee is `@ExperimentalSerializationApi`, which isn't used here.

I also verified it against this branch:

- The full `serializer-deserializer` test suite passes: 1332 tests, 0 failures, 0 errors, 0 skipped (JDK 21, Maven 3.9).
- `mvn dependency:tree` resolves `kotlinx-serialization-core-jvm:1.7.3` cleanly with no version conflict or mediation.

If anything, this reduces breakage: the pinned 1.4.0 was forcing a downgrade for downstream users already on kotlinx.serialization 1.6+, producing the "current kotlinx.serialization core version is too low" error (#313). 1.7.3 aligns with the Kotlin 1.9.25 / Wire 5.2.0 versions already in use.

subramp (approved):

isaurab007:
The change looks fine to me and so does the reasoning behind backwards compatibility. Are we able to attach integration and unit test evidence?

awsmasudur:
Yes — I ran the Kafka integration tests locally against **real AWS Glue Schema Registry** on this branch (kotlinx-serialization-core-jvm `1.7.3`, commit `fdc64ce`). Since this bump is a transitive dependency for Wire's protobuf support, I focused on exercising the protobuf serialize → register → deserialize round-trip end-to-end.

**Setup**
- Branch: `fix/issue-313-bump-kotlinx-serialization` (kotlinx-serialization-core-jvm = 1.7.3)
- Kafka: local broker on `127.0.0.1:9092`
- Schema Registry: **real** AWS Glue, region `us-east-2`
- Command: `mvn --file integration-tests/pom.xml verify -Psurefire -Dtest=GlueSchemaRegistryKafkaIntegrationTest`

**Result — all green**

```
[INFO] Tests run: 45, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 999.442 s
        - in ...integrationtests.kafka.GlueSchemaRegistryKafkaIntegrationTest
[INFO] Tests run: 45, Failures: 0, Errors: 0, Skipped: 0
...
[INFO] BUILD SUCCESS
[INFO] Total time:  33:39 min
```

(The `-Psurefire` profile binds the class to both the `integration-test` and `verify` phases, so it executed twice — both runs reported 45/45 passing.)

**Protobuf coverage specifically**

The run auto-registered **52 PROTOBUF schemas** in Glue across NONE and BACKWARD compatibility, e.g.:
- `SchemaRegistryTests-…-PROTOBUF-NONE`
- `SchemaRegistryTestsStreamsInput-…-PROTOBUF-BACKWARD`

Sample protobuf produce/consume round-trip from the logs:
```
Auto Creating schema … PROTOBUF-NONE … message Address { required string street = 1; optional int32 zip = 2; optional string city = 3; } …
KafkaHelper - Received message 2: key = message-0, value = street: "Foo"
```

So protobuf serialization/deserialization via Wire works correctly end-to-end against real Glue with kotlinx-serialization 1.7.3. The parameterized suite also covers AVRO and JSON across the record-type / compatibility / compression matrix, all passing.

Happy to share the full log or run any additional scenario if useful.

knoxy5467:
If the compatibility contract is valid for backwards compatibility between minor versions then this PR would not be needed since 1.4 -> 1.7 is a minor version bump. Since this PR is being raised because of a compatibility problem it directly points to the need to create a compatibility test with a project using 1.4 consuming this package before and after the kotlin versions bump. Additionally if the compatibility contract should hold for all minor versions then we should be bumping to the latest instead of 1.7. 

Please run the integration tests and also an exploratory test showing backwards compatibility with an "existing" project using 1.4 that consumes this package on both the current and new dependency graphs.

subramp:
Thanks for picking up #313. I dug into this while evaluating what a newer version would look like for the 2.0.0 major, and ended up at a different fix, which I have opened as #542.

The short version: the version number is not the mechanism, the **scope** is. `serializer-deserializer/pom.xml` declares `kotlinx-serialization-core-jvm` at compile scope, which puts the jar on every downstream consumer's `compileClasspath`, where the Kotlin serialization compiler plugin reads `Require-Kotlin-Version` from the manifest and hard-fails on older compilers. Bumping to 1.7.3 keeps it at compile scope, and 1.7.3 is the first release that declares `Require-Kotlin-Version: 2.0.0-RC1`, so it breaks compilation for Kotlin 1.9.x consumers, including the 1.9.25 that this repo itself is pinned to. Verified from the published jar manifests:

| core version | `Require-Kotlin-Version` |
|---|---|
| 1.4.0 (current) | 1.4.30-M1 |
| 1.6.3 | 1.4.30-M1 |
| 1.7.3 (this PR) | 2.0.0-RC1 |
| 1.11.0 (latest) | 2.0.0-RC1 |

A second thing surfaced while looking at this: the explicit declaration is currently *downgrading* what Wire ships. `wire-compiler:5.2.0` asks for core 1.7.3 at runtime scope, but the hand-written 1.4.0 wins Maven nearest-wins, so `kaml-jvm:0.67.0` runs against a core it was not built against. The same pin holds `kotlin-stdlib` at 1.9.25 while Wire 5.2.0 asks for 2.0.21.

Since GSR has zero source references to `kotlinx-serialization-core` (it is needed only at runtime by Wire's kaml), #542 deletes the declaration rather than bumping it and lets `wire-compiler` supply it transitively at runtime scope. That takes it off the consumer compile classpath permanently, at any version. It also moves Wire 5.2.0 to 7.0.1, which resolves core 1.11.0 and kaml 0.104.0 on its own, plus the `kotlin.version` and okio alignment that requires and one source fix for the `weakImports` parameter added to `ProtoFileElement` in Wire 7.

Test results on #542: 1346/1346 unit tests in `serializer-deserializer`, and 74/74 integration tests against real Glue Schema Registry (`mvn --file integration-tests/pom.xml verify -Psurefire`), covering Kafka 45/45, Kinesis 28/28, and the protobuf converter 1/1.

Because the stdlib bump raises the Kotlin floor for consumers, #542 targets the 2.0.0 major rather than a patch. Happy to defer to you on how you would like to handle this one, whether that is closing this in favour of #542 or folding the scope change in here instead.

awsmasudur:
Thanks @subramp, this is a great root-cause analysis and it's convincing. You're right that the scope is the real mechanism, not the version. The `compile`-scope declaration is what puts `kotlinx-serialization-core-jvm` on every consumer's compile classpath, where the serialization compiler plugin reads `Require-Kotlin-Version` from the manifest and fails on older compilers.

The manifest table is the deciding point: 1.7.3 is the first release requiring `Require-Kotlin-Version: 2.0.0-RC1`, so bumping to it would break compilation for Kotlin 1.9.x consumers — including the 1.9.25 this repo is pinned to. So this PR's approach isn't just insufficient, it would introduce a new break. And as you noted, the explicit pin was also silently downgrading what Wire ships at runtime, which is its own problem.

Removing the declaration entirely and letting `wire-compiler` supply it transitively at runtime scope (plus the Wire 5.2.0 → 7.0.1 move) is the correct fix — it takes the dependency off the consumer compile classpath permanently, at any version, rather than chasing version numbers. That also addresses @knoxy5467's point: once it's off the compile classpath, the minor-version compatibility question is moot.

Since #542 has already been merged and resolves #313 the right way, I'll **close this in favor of #542**. Thanks @subramp for digging in, and thanks @knoxy5467 and @isaurab007 for the scrutiny that pushed this toward the correct fix.

PR #532, opened 2026-07-24
fix(protobuf): Handle STRUCT without metadata

### LINKED ISSUE #289: ConnectSchemaToProtobufSchemaConverter throws NullPointerException
I'm using the Debezium connector with ProtobufSchemaConverter as the value converter and auto-registering schema in Glue. It's unclear what is actually the source of the null value.

```
Caused by: java.lang.NullPointerException
  at com.amazonaws.services.schemaregistry.kafkaconnect.protobuf.fromconnectschema.FieldBuilder.build(FieldBuilder.java:92)
  at com.amazonaws.services.schemaregistry.kafkaconnect.protobuf.fromconnectschema.ConnectSchemaToProtobufSchemaConverter.buildMessageDescriptorProto(ConnectSchemaToProtobufSchemaConverter.java:81)
  at com.amazonaws.services.schemaregistry.kafkaconnect.protobuf.fromconnectschema.ConnectSchemaToProtobufSchemaConverter.buildFileDescriptorProto(ConnectSchemaToProtobufSchemaConverter.java:66)
  at com.amazonaws.services.schemaregistry.kafkaconnect.protobuf.fromconnectschema.ConnectSchemaToProtobufSchemaConverter.convert(ConnectSchemaToProtobufSchemaConverter.java:41)
  at com.amazonaws.services.schemaregistry.kafkaconnect.protobuf.ProtobufSchemaConverter.fromConnectData(ProtobufSchemaConverter.java:105)
  at org.apache.kafka.connect.storage.Converter.fromConnectData(Converter.java:63)
  at org.apache.kafka.connect.runtime.WorkerSourceTask.lambda$convertTransformedRecord$3(WorkerSourceTask.java:321)
  at org.apache.kafka.connect.runtime.errors.RetryWithToleranceOperator.execAndRetry(RetryWithToleranceOperator.java:156)
  at org.apache.kafka.connect.runtime.errors.RetryWithToleranceOperator.execAndHandleError(RetryWithToleranceOperator.java:190
```

a-lmillette:
*Issue #, if available:*

Fixes #289

*Description of changes:*

Kafka Connect schemas may return `null` from `Schema.parameters()` when no
parameters are defined. This occurs with STRUCT fields produced by Debezium
that do not contain protobuf-specific metadata.

Add a null check before reading oneof metadata in `FieldBuilder`. STRUCT fields
without metadata now convert as nested protobuf messages instead of throwing a
`NullPointerException`. Existing oneof conversion behavior is unchanged.

*Testing:*

Ran `ConnectSchemaToProtobufSchemaConverterTest` under JDK 11:

- Tests run: 14
- Failures: 0
- Errors: 0

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.


knoxy5467:
Hello! can you please run the integration tests in the repository and provide them as a part of the PR? they are not ran during the github workflows due to requiring a real AWS account.

Additionally can you create and run some tests targeted at this change to confirm that it is not a backwards incompatible change for both the Java and the Csharp libraries?

a-lmillette:
## Additional testing and backward-compatibility validation

Testing uncovered a second occurrence of the null-parameters issue. The original
change allowed schema construction to succeed, but full record conversion still
dereferenced `Schema.parameters()` in
`ConnectDataToProtobufDataConverter`. I added the matching null guard there and
hardened both oneof checks against null metadata values.

### Backward compatibility

The exact triggering input cannot be produced by the old implementation because
it fails before publishing a record. To validate wire compatibility, I compared:

- The previously supported nested STRUCT representation with an empty parameter
  map.
- The newly supported representation where `Schema.parameters()` returns null.

Both produce identical Protobuf descriptors and identical Protobuf payload
bytes, and the nested field value is preserved. Therefore, the fix does not
introduce a new wire representation; it enables the existing ordinary nested
message representation for a previously failing input.

No public APIs, configuration keys, defaults, compatibility modes, consumer
logic, or schema-evolution behavior changed.

### Java results

- Full JDK 11 Maven build: PASS
- Clean Protobuf Kafka Connect converter reactor:
  - Tests: 96
  - Failures: 0
  - Errors: 0
- Targeted live Glue metadata-free STRUCT round trip:
  - Tests: 1
  - Failures: 0
  - Errors: 0
- Full Java integration profile:
  - Tests: 74
  - Failures: 0
  - Errors: 0
  - Covered Kafka and Kinesis with Avro, JSON, and Protobuf

### C# results

The changed Kafka Connect converter is not directly used by the C# binding, but
I ran the downstream C# regression and live Glue suites to confirm no propagated
regression:

- Linux ARM64 native GraalVM/C/SWIG/.NET build: PASS
- CI-filtered serializer/deserializer unit suite:
  - Tests: 88
  - Failures: 0
- Live Glue configuration and schema-evolution suites:
  - Tests: 22
  - Failures: 0

### Resource cleanup

All test-created schemas were deleted. Post-test audits in `us-east-1`,
`us-east-2`, and `us-west-2` found no remaining test schemas or custom test
registries.

protobuf-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/protobuf/fromconnectdata/ConnectDataToProtobufDataConverter.java
```diff
@@ -55,8 +55,9 @@ public Message convert(
             if (field.schema().type().equals(Schema.Type.MAP)) {
                 addMapField(fileDescriptor, dynamicMessageBuilder, field, fieldValue);
             } else if (field.schema().type().equals(Schema.Type.STRUCT)) {
-                if (field.schema().parameters().containsKey(PROTOBUF_TYPE)
-                        && field.schema().parameters().get(PROTOBUF_TYPE).equals(PROTOBUF_ONEOF_TYPE)) {
+                if (field.schema().parameters() != null
```
subramp: nit: `field.schema().parameters()` is evaluated three times in this condition. Consider hoisting to a local for readability:

```
 final Map<String, String> params = field.schema().parameters();
 if (params != null && PROTOBUF_ONEOF_TYPE.equals(params.get(PROTOBUF_TYPE))) {
```

protobuf-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/protobuf/fromconnectschema/FieldBuilder.java
```diff
@@ -89,8 +89,9 @@ public static void build(
                 messageDescriptorProtoBuilder.addNestedType(buildMap(fieldSchema, mapEntryName,
                     fileDescriptorProtoBuilder, messageDescriptorProtoBuilder));
             } else if (Schema.Type.STRUCT.equals(fieldSchema.type())) {
-                if (fieldSchema.parameters().containsKey(PROTOBUF_TYPE)
-                        && fieldSchema.parameters().get(PROTOBUF_TYPE).equals(PROTOBUF_ONEOF_TYPE)) {
+                if (fieldSchema.parameters() != null
```
subramp: nit: Similar to above,  three `fieldSchema.parameters()` calls in one condition; a local would read cleaner.

protobuf-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/protobuf/fromconnectdata/ConnectDataToProtobufDataConverter.java
a-lmillette: Addressed in ad99c87 by hoisting the schema parameters to a local variable and simplifying the oneof check. Thanks!

protobuf-kafkaconnect-converter/src/main/java/com/amazonaws/services/schemaregistry/kafkaconnect/protobuf/fromconnectschema/FieldBuilder.java
a-lmillette: Addressed in ad99c87 by hoisting the schema parameters to a local variable and simplifying the oneof check. Thanks!

subramp (approved):

jvdadda:
Hi @a-lmillette, thanks for this one, and for the follow-up of 1 September in particular: the second dereference on the record conversion path was a real gap, and our maintained fork of the Java library had it too. Your first fix (the `FieldBuilder` null guard, plus a fallback name for an unnamed struct) is in the fork as https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/67; the `ConnectDataToProtobufDataConverter` site you found is fixed in https://github.com/mobsuccess-devops/aws-glue-schema-registry/pull/176, credited to your report, with a test converting a full record rather than just the schema. Both are available in the latest published release, currently 4.1.0, on Maven Central as `com.mobsuccess:protobuf-kafkaconnect-converter`; Connect workers take the `-plugin.zip` attached to the release.

One edge we know is still open on our side, in case you hit it with Debezium: an unnamed STRUCT nested inside an array or a map still fails while the schema is built, before either guard runs.

isaurab007 (approved):

PR #533, opened 2026-08-04
feat(json): gate className resolution behind an explicit class allowlist

subramp:
Fix broken tests and add allowlist to narrow opt-in flag scope

jsonClassNameResolutionEnabled=true restores the legacy behavior of
resolving the schema's className via Class.forName and deserializing
into that POJO. That re-opens the original concern for anyone who needs
typed deserialization: the schema, not the consumer, still picks which
class gets instantiated, so a schema an attacker can influence can name
any class on the classpath.

Add jsonClassNameAllowlist, a comma-separated list of fully qualified
class names the deserializer may instantiate. Resolution now requires
both the flag and a matching allowlist entry; a className absent from
the list deserializes to JsonDataWithSchema, exactly as if resolution
were off. The allowlist defaults to empty, so the flag alone is inert
and consumers name the classes they actually expect on the topic. The
javadoc and CHANGELOG are updated to say both properties are required,
since the flag on its own no longer changes anything.

The allowlist miss is a configuration-scoped condition, not a
record-scoped one, so logging it per record would emit at message
throughput rate. warnOnceForDisallowedClassName dedups by class name via
a concurrent set, excluded from Lombok's equals/hashCode/toString so
that having logged does not make two deserializer instances unequal.

GlueSchemaRegistryDeserializationFacadeTest was not updated for the
changed default, so six cases were failing: the JSON specific-record
cases of testDeserialize_withValidSchemaResponse_recordMatches and
testDeserialize_withSerdeConfigs_recordMatches assert a typed POJO comes
back from a schema carrying a className, which no longer happens by
default. They now opt in and allowlist that class.

testDeserialize_invalidSpecificJsonRecord_throwsException is the subtle
one: it asserts a bogus className throws, but with resolution disabled
it would pass for the wrong reason, because resolution is never
attempted. It now allowlists the bogus name so Class.forName remains
what fails.

Tests: allowlist parsing (comma-separated, whitespace-tolerant, empty
default) in GlueSchemaRegistryConfigurationTest; allowlist hit, miss,
empty, and multi-entry cases in JsonDeserializerTest, plus a repeated
record case asserting the warn-once cache does not alter what later
records deserialize to.


knoxy5467 (commented):
Reviewed the full diff. This is a well-motivated, secure-by-default hardening — the "flag alone is inert" design, the null-safe gating, the correct Lombok exclusions on the mutable warn-tracking state, and the test matrix (enabled/disabled × allowlisted/miss/empty/multi-entry, plus the repeated-record warn-once test) are all good. The `INVALID_CLASS_NAME = "wrong.class.name"` allowlisting so that `Class.forName` remains the failing step is a genuinely subtle, correct fix. Version bump 1.1.27 → 1.1.28 is consistent across all poms and the docs (CHANGELOG/README/javadoc) are thorough.

Leaving 4 inline notes: one worth addressing before merge (#1, unbounded warn-set under the PR's own threat model), the rest are polish. No blocking issues.

serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/json/JsonDeserializer.java
```diff
@@ -109,4 +130,20 @@ public Object deserialize(@NonNull ByteBuffer buffer,
...
+
+    /**
+     * Warns that a schema's className was not allowlisted, at most once per distinct class name.
+     * The condition is configuration-scoped rather than record-scoped, so logging it on every
+     * record would flood the logs at message throughput rate.
+     *
+     * @param className the class name named by the schema but absent from the allowlist
+     */
+    private void warnOnceForDisallowedClassName(String className) {
+        if (warnedClassNames.add(className)) {
```
knoxy5467, line 176: **`warnedClassNames` can grow unbounded on attacker-controlled input.** The dedup key is `className`, which comes straight from the schema. This PR's own threat model states a schema an attacker can influence "can name any class on the classpath." Under that same model, a stream of records carrying many *distinct* disallowed class names grows this set without bound for the life of the deserializer instance — a slow memory-growth / DoS vector, somewhat ironic given the security intent. In ordinary misconfiguration it's a handful of entries, so severity is low-to-moderate, but worth bounding since the feature is explicitly adversarial.

Suggestion: cap the set (e.g. only add below a max size, accepting occasional duplicate warns above the cap):
```java
if (warnedClassNames.size() < MAX_WARNED_CLASS_NAMES && warnedClassNames.add(className)) {
    log.warn(...);
}
```

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
```diff
@@ -270,6 +277,24 @@ private void validateAndSetSchemaAutoRegistrationSetting(Map<String, ?> configs)
...
+                            .toString());
+        } else {
+            log.info("jsonClassNameResolutionEnabled is not defined in the properties. Using the default value {}",
+                     jsonClassNameResolutionEnabled);
+        }
+
+        if (isPresent(configs, AWSSchemaRegistryConstants.JSON_CLASS_NAME_ALLOWLIST)) {
+            String allowlistValue = configs.get(AWSSchemaRegistryConstants.JSON_CLASS_NAME_ALLOWLIST).toString().trim();
+            if (!allowlistValue.isEmpty()) {
+                this.jsonClassNameAllowlist = new HashSet<>(Arrays.asList(allowlistValue.split("\\s*,\\s*")));
```
knoxy5467: **Leading/doubled commas inject an empty-string entry into the allowlist.** `",com.example.Foo"` or `"a,,b"` splits to include a `""` element. It's functionally harmless (an empty string never matches a real class name), but it's junk in the set and slightly surprising. Consider filtering blanks:
```java
this.jsonClassNameAllowlist = Arrays.stream(allowlistValue.split("\\s*,\\s*"))
        .filter(s -> !s.isEmpty())
        .collect(Collectors.toSet());
```

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
```diff
@@ -270,6 +277,24 @@ private void validateAndSetSchemaAutoRegistrationSetting(Map<String, ?> configs)
         }
     }
 
+    private void validateAndSetJsonClassNameResolutionSetting(Map<String, ?> configs) {
+        if (isPresent(configs, AWSSchemaRegistryConstants.JSON_CLASS_NAME_RESOLUTION_ENABLED)) {
+            this.jsonClassNameResolutionEnabled = Boolean.parseBoolean(
```
knoxy5467: Minor / consistency: a malformed value like `"ture"` or `"yes"` silently becomes `false` via `Boolean.parseBoolean`, so a consumer who fat-fingers the flag believes resolution is on when it's off. This matches the existing `validateAndSet...` convention for the other boolean flags in this file, and for a *security* opt-in defaulting a bad value to "off" is at least the safe direction — so no change strictly required. Flagging only for awareness; a WARN on an unrecognized value would be friendlier.

serializer-deserializer/src/test/java/com/amazonaws/services/schemaregistry/deserializers/GlueSchemaRegistryDeserializationFacadeTest.java
```diff
@@ -874,6 +889,25 @@ private byte[] createSerializedCompressedEmployeeData(Object objectToSerialize,
...
+     */
+    private void enableJsonClassNameResolution(String allowedClassName) {
+        configs.put(AWSSchemaRegistryConstants.JSON_CLASS_NAME_RESOLUTION_ENABLED, true);
+        configs.put(AWSSchemaRegistryConstants.JSON_CLASS_NAME_ALLOWLIST, allowedClassName);
+    }
+
+    /**
+     * Restores the default (disabled) className resolution behavior for subsequent tests.
+     */
+    private void disableJsonClassNameResolution() {
```
knoxy5467: Minor / matches existing pattern: `enableJsonClassNameResolution(...)` mutates the shared `configs` field and the paired `disableJsonClassNameResolution()` runs at the end of the test body rather than in a `finally`, so a mid-test assertion failure skips cleanup. This is **safe today** because JUnit 5's default (PER_METHOD) lifecycle rebuilds the test instance — and thus `configs` — for every test, and it's consistent with the file's existing `configs.put(...)/configs.remove(...)` style. Noting only because the pattern would silently leak state if anyone ever adds `@TestInstance(PER_CLASS)`.

pom.xml
```diff
@@ -21,7 +21,7 @@
 
     <groupId>software.amazon.glue</groupId>
     <artifactId>schema-registry-parent</artifactId>
-    <version>1.1.27</version>
+    <version>1.1.28</version>
```
knoxy5467: **Blocking: a breaking change requires a major version bump, not a patch bump.** This is `1.1.27 → 1.1.28` (patch), but the CHANGELOG classifies this release as a breaking change — by default the JSON deserializer now returns `JsonDataWithSchema` instead of the resolved POJO, so consumers that relied on automatic POJO deserialization will silently start failing on the cast.

Under semver, a backwards-incompatible change bumps the **major** version. A consumer on a `1.1.x` range (a common Maven / Dependabot setup) would pick this up as a routine patch and break with no version signal. Please bump to **`2.0.0`** across all the poms instead of `1.1.28` (and update the CHANGELOG/README `1.1.28` references to match).

If the intent is deliberately *not* to follow strict semver here (this repo's `1.1.x` history hasn't always), that's a maintainer call — but then it should be explicit, because shipping a documented breaking change under a patch bump is the case semver exists to prevent.

knoxy5467 (changes requested):
Requesting changes on the version bump. This PR ships a change the CHANGELOG itself labels a breaking change ("JSON deserializer no longer resolves the schema's `className` field into a POJO by default... This is a breaking change for consumers that relied on automatic POJO deserialization"), but the version only moves 1.1.27 → 1.1.28, a patch bump. Under semver a backwards-incompatible change should bump the major version. See the inline comment on the root pom for the specifics and requested fix.

serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/json/JsonDeserializer.java
subramp: Capped at `MAX_WARNED_CLASS_NAMES = 100`, short-circuiting on size before the add ([JsonDeserializer.java:153](vscode-webview://0nsffpnt7j2cstpeq90bjrulplsp3rjj8ski37u2tjf8vrdqt3p7/serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/json/JsonDeserializer.java#L153)):

```
if (warnedClassNames.size() >= MAX_WARNED_CLASS_NAMES || warnedClassNames.add(className)) {
```
Warns every time past the cap rather than going silent, dropping the warning would hide a real misconfiguration exactly when it matters. Duplicate log lines are the cheaper failure mode. `testDeserialize_manyDistinctDisallowedClassNames_warnStateStaysBounded` feeds 500 distinct names and asserts the bound.

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
subramp:
```
Set<String> allowedClassNames = Arrays.stream(allowlistValue.split(","))
        .map(String::trim)
        .filter(className -> !className.isEmpty())
        .collect(Collectors.toSet());
```

Covered by `testJsonClassNameAllowlist_strayCommas_areIgnored` and `..._onlySeparators_defaultsToEmpty`.

subramp (commented):
Ready for revision 2 review

serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/json/JsonDeserializer.java
```diff
@@ -43,7 +49,17 @@
...
+     * Upper bound on {@link #warnedClassNames}. The dedup key comes from the schema, which a
+     * producer controls, so an unbounded set would grow for the lifetime of the deserializer.
+     * Past this many distinct class names the warning is logged every time instead of once.
+     */
+    static final int MAX_WARNED_CLASS_NAMES = 100;
     private final ObjectMapper objectMapper;
+    /** Class names already warned about, so the allowlist-miss warning is not logged per record. */
+    @EqualsAndHashCode.Exclude
+    @ToString.Exclude
+    private final Set<String> warnedClassNames = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
isaurab007, line 69: Why do we want to warn only once for 100 classes and once per record beyond this limit? Why not something arbitrarily big like the total number of schemas allowed in an AWS account per region? This makes sure there are no excessive warnings but also battles the unbounded growth of  `warnedClassNames` problem.

Also would you be elaborate in the comment what are the scenarios where we `warn` and comment on are we making a change to warning behavior in the same PR where we are introducing a class allowlist.

README.md
```diff
@@ -279,8 +279,10 @@ The recommended way to use the AWS Glue Schema Registry Library for Java is to c
 You could use a Java POJO and pass the object as a record.
 We use [mbknor-jackson-jsonschema](https://github.com/mbknor/mbknor-jackson-jsonSchema) to generate a JSON Schema for
  the POJO passed. This library can also inject additional information in the JSON Schema.
- 
- **GSR Library uses the "className" to fully classified class name to deserialize back to an Object of the POJO**
+
+ **GSR Library uses the "className" fully qualified class name to deserialize back to an Object of the POJO.
+ As of 2.0.0 this is opt-in — see [Deserializing JSON into a Java POJO (className resolution)](#deserializing-json-into-a-java-pojo-classname-resolution).
```
isaurab007: Nit: Instead of saying 'This' is opt-in and without 'it', can we expand that this is a new feature added in 2.0.0 and use its name for enhanced clarity?

README.md
```diff
@@ -339,6 +341,37 @@ public class Car {
 
 ```
 
+### Deserializing JSON into a Java POJO (className resolution)
+
+By default the JSON deserializer returns a `JsonDataWithSchema`, even when the schema carries a
+`className` property. Resolving that property would let the schema decide which class the
+deserializer instantiates via reflection, so it must be opted into explicitly.
```
isaurab007, line 349: Why does line 284 say `as of 2.0.0 this is opt-in` when we say here that the customers need to opt-in explicitly?

serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/json/JsonDeserializer.java
```diff
@@ -94,9 +110,20 @@ public Object deserialize(@NonNull ByteBuffer buffer,
...
 
-            if (classNameNode != null) {
+            boolean classNameResolutionEnabled = schemaRegistrySerDeConfigs != null
+                    && schemaRegistrySerDeConfigs.isJsonClassNameResolutionEnabled();
+
+            if (classNameResolutionEnabled && classNameNode != null) {
                 String className = classNameNode.asText();
-                deserializedObject = objectMapper.readValue(data, Class.forName(className));
+                Set<String> allowlist = schemaRegistrySerDeConfigs.getJsonClassNameAllowlist();
+                if (allowlist != null && allowlist.contains(className)) {
```
isaurab007: I am not sure if we want to support regex matches here instead of full-string matches which are harder for customers to setup if they want to just allow all the classes.

subramp (commented):
Addressed all review comments isaurab007

serializer-deserializer/src/main/java/com/amazonaws/services/schemaregistry/deserializers/json/JsonDeserializer.java
subramp: On the quota as the cap: they measure different things. The cap bounds distinct disallowed class names one consumer sees, a handful in practice. The account quota bounds schemas registered across the account, and doesn't upper-bound the set anyway, since one schema can be updated with a new className at any time. Sizing the cap to the quota puts it orders of magnitude above anything a real consumer reaches.

On "once per 100, then once per record": agreed, that trade was wrong and it's gone. Per-record warning past the cap just swaps a memory flood for a log flood. It now logs once on reaching the cap that it's suppressing further warnings, then stops. By then 100 names are logged, so the problem is already surfaced.

On documenting when we warn: the warnOnceForDisallowedClassName javadoc now states the scenario (resolution enabled, className present, not allowlisted) and the suppression behavior. This isn't a warning bolted onto an unrelated feature: the allowlist-miss warning is the allowlist. It didn't exist before this PR, so warning and bounding that warning are the same change. The only pre-existing thing touched is the README's "once per distinct class name" line, which the cap made false, now corrected.

README.md
subramp: They say the same thing. "Opt-in as of 2.0.0" is the one-line summary at the POJO section; "customers need to opt in explicitly" is the same rule stated in the detailed section it links to. Opt-in means resolution is off until the customer turns it on, which is exactly the explicit step the linked section describes (set jsonClassNameResolutionEnabled and populate the allowlist). Line 284 is the pointer, the linked section is the how.

subramp (commented):
Responded to comments

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
```diff
@@ -294,19 +300,107 @@ private void validateAndSetJsonClassNameResolutionSetting(Map<String, ?> configs
...
             // never-matching "" into the allowlist.
-            Set<String> allowedClassNames = Arrays.stream(allowlistValue.split(","))
+            Set<String> allowedClassNames = rawEntries
                     .map(String::trim)
                     .filter(className -> !className.isEmpty())
                     .collect(Collectors.toSet());
+            // A bare "*" would allow every class on the classpath, which is the behavior this
+            // allowlist exists to prevent. Reject it rather than honoring it, so the opt-in cannot
+            // be widened back to the pre-2.0.0 default by a single character.
+            if (allowedClassNames.contains("*") || allowedClassNames.contains(PACKAGE_WILDCARD_SUFFIX)) {
```
isaurab007, line 326: Nit: I am not sure if we want to prevent wildcard entirely or just make sure customer is intentional whenever they set it (which the feature flag already ensures). But for now, I am okay leaving it as is.

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
```diff
@@ -294,19 +300,107 @@ private void validateAndSetJsonClassNameResolutionSetting(Map<String, ?> configs
...
+     */
+    private static boolean isBareWildcard(String entry) {
+        return "*".equals(entry) || PACKAGE_WILDCARD_SUFFIX.equals(entry);
+    }
+
+    /**
+     * Whether {@code className} sits directly in the package named by a {@code ".*"} allowlist
+     * entry, with no further package segment. The remainder after the package prefix must be a
+     * simple class name, so {@code com.example.pojos.nested.Car} does not match
+     * {@code com.example.pojos.*}. Nested classes such as {@code Car$Engine} do match, as they
```
isaurab007, line 385: I am trying to understand if we are giving enough flexibility here if we are explicitly opposing nested class paths. What if someone wants to do `com.foo.company.*`

README.md
isaurab007: Nit: I interpret `Opt-in as of 2.0.0` as opt-in by default. Can we instead use 'Introduced in 2.0.0; disabled by default"

isaurab007 (approved):
Looks good, but I have one critical comment where we need to rethink the regex pattern matching - whether we want to support multi-level nesting to allow more flexibility.

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
subramp: The flag makes enabling resolution intentional, but a bare `*` is a second, sharper footgun: it re-opens the `pre-2.0.0 `"resolve any class" behavior in one character, inside a config a producer's tooling might template. Rejecting it keeps "enabled" from silently meaning "everything." Happy to revisit if it proves annoying in practice.

common/src/main/java/com/amazonaws/services/schemaregistry/common/configs/GlueSchemaRegistryConfiguration.java
subramp: Deliberately direct-members-only. `*` already auto-trusts classes added to that package later, which is a bounded, visible concession; `**` would extend that to a whole subtree, so a new sub-package of instantiable classes would get trusted with no allowlist edit. That's the widest walk-back of the explicit-allowlist property this change exists to add. For a hierarchy, a line per package (`com.foo.company.orders.*`, ..`.billing.*`) stays explicit and is cheap, so I'd rather not add a recursive form.

subramp (commented):
Minor readme update

isaurab007 (approved):

knoxy5467 (approved):
Please run the integration tests. Otherwise looks good.

PR #536, opened 2026-08-18
security: pin Actions to SHAs, add Dependabot, add zizmor CI

knoxy5467:
## What this does

Hardens the GitHub Actions supply chain for this repo:

1. **Pin all third-party Actions to full-length commit SHAs.**
   Every `uses:` in `MavenCI.yml` and `build-csharp.yml` now uses a 40-character SHA plus a comment naming the tag it resolves. Tag-based pins can be moved; SHA pins cannot. This defends against a compromised or retagged upstream Action.

2. **Add Dependabot for three ecosystems.**
   `.github/dependabot.yml` covers `github-actions` at `/`, `maven` at `/`, and `nuget` at `/multilang-schema-registry/csharp/AWSGsrSerDe`. Dependabot will now open weekly bump PRs so pinned SHAs do not go stale.

3. **Add zizmor CI.**
   `.github/workflows/zizmor.yml` runs zizmor on every push and pull request that touches `.github/workflows/**`. It reports findings to the Code Scanning tab. This catches new workflow issues before they land.

4. **Least-privilege plus no-persist-credentials.**
   Both existing workflows now declare top-level `permissions: contents: read` and every `actions/checkout` sets `persist-credentials: false`. The zizmor workflow uses `permissions: {}` at the top and grants `security-events: write` only on the one job that needs it.

## Why

CI is often the softest target in a repo. Unpinned Actions can be swapped upstream, over-scoped `GITHUB_TOKEN` can be reused across steps, and un-audited workflow YAML tends to drift. This PR closes those three gaps.

## What is not in this PR

- No changes to build, test, or release logic.
- No changes to code outside `.github/`.
- The pinned Action major versions are unchanged in this PR. The follow-up bumps (for example `actions/checkout@v2` and `actions/setup-java@v1`) will come as separate Dependabot PRs.

## How to verify

- `gh api repos/<owner>/<action>/git/ref/tags/<tag>` on each pinned SHA confirms the SHA matches the tag its comment claims.
- Running `zizmor .github/workflows/` locally on this branch prints no findings.
- The workflow itself will run on this PR and post any findings to the Checks tab.

## Note on zizmor's `known-vulnerable-actions` check

zizmor has one online audit, `known-vulnerable-actions`, that calls the GitHub API and needs a token. A local run without a token skips that audit. The CI run on this PR runs with `GH_TOKEN` and does exercise it, so the Checks tab on this PR is the real gate for that audit.

## Compatibility

Backwards-compatible. No public API, published artifact, or user-facing behavior changes.

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -1,0 +1,31 @@
+name: Check GitHub Actions
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - "master"
+  pull_request:
+    branches:
+      - "master"
+
+permissions: {}
+
+jobs:
+  zizmor:
+    name: Check GitHub Actions security
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@6fc4b006235f201fdab3722e17240ab420d580e5 # v0.6.1
+        with:
+          advanced-security: ${{ github.event_name == 'push' && 'true' || 'false' }}
+          min-severity: low
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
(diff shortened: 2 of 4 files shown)


.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Checkout
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@6fc4b006235f201fdab3722e17240ab420d580e5 # v0.6.1
+        with:
+          advanced-security: ${{ github.event_name == 'push' && 'true' || 'false' }}
+          min-severity: low
```
isaurab007, line 31: Making sure this is the right sensitivity. Did not find any zizmor findings in checks tab, so moving ahead with approval.

isaurab007 (approved):

PR #538, opened 2026-09-02
build: remove AWS SDK v1 (aws-java-sdk-sts) and migrate remaining v1 usages to SDK v2

cmakineni:
# build: remove AWS SDK v1 (`aws-java-sdk-sts`) and migrate remaining v1 usages to SDK v2

## What & why

AWS SDK for Java v1 is end-of-support. The published modules already use SDK v2 exclusively — the only v1 left was an unused `aws-java-sdk-sts` dependency plus a few usages confined to the non-published `examples` and integration-test code. This removes v1 from those places and drops the v1 dependency management from the parent POM. Since no published source referenced a v1 class, removing `aws-java-sdk-sts` doesn't change library behavior.

## Changes

- **`serializer-deserializer`** — drop the unused `com.amazonaws:aws-java-sdk-sts` (v2 `software.amazon.awssdk:sts` is already used); rename `AWSCredentialsProvider` → `AwsCredentialsProvider` in Javadoc/comments.
- **`multilang-schema-registry`** — drop the unused `com.amazonaws:aws-java-sdk-sts`.
- **`examples`** — migrate `PutRecordGetRecordExample` from v1 `AmazonKinesis` to v2 `KinesisClient` (`aws-java-sdk-kinesis` → `software.amazon.awssdk:kinesis`).
- **`integration-tests`** — resolve the test region via v2 `DefaultAwsRegionProviderChain` instead of v1 `com.amazonaws.regions.Regions`.
- **parent `pom.xml`** — remove the now-unused `aws.sdk.v1.version` property and the `aws-java-sdk-kinesis` dependency-management entry.

No `com.amazonaws:aws-java-sdk-*` artifact remains in the build.

## Backwards compatibility

- No public API, config, wire-format, default, or error-behavior change; no behavior-contract box checked.
- Only effect is transitive: consumers still using AWS SDK v1 must now declare `aws-java-sdk-sts` themselves. `examples` and `integration-tests` are not published, so their migrations have no consumer impact.

## Tests

- **Full parent build** (`mvn clean verify`): all 13 modules green — unit tests, checkstyle, jacoco ≥85%.
- **Integration** (real Glue, us-east-2; not run in CI): `verify -Psurefire` — **73/0** (Kafka 45, Kinesis 28).


integration-tests/src/test/java/com/amazonaws/services/schemaregistry/integrationtests/properties/GlueSchemaRegistryConnectionProperties.java
```diff
@@ -13,10 +13,20 @@
...
 package com.amazonaws.services.schemaregistry.integrationtests.properties;
-import com.amazonaws.regions.Regions;
+
+import software.amazon.awssdk.core.exception.SdkClientException;
+import software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain;
 
 public interface GlueSchemaRegistryConnectionProperties {
     // Glue Service Endpoint
-    String REGION = Regions.getCurrentRegion() == null ? "us-east-2" : Regions.getCurrentRegion().getName().toLowerCase();
+    String REGION = resolveRegion();
```
subramp, line 25: `ENDPOINT` still hardcodes .amazonaws.com while `REGION` is now ambient. Pre-existing, but this PR makes it reachable: if the chain resolves `cn-north-1` or `us-gov-west-1`, the endpoint becomes `https://glue.cn-north-1.amazonaws.com`, which is wrong and fails confusingly. `ServiceMetadata.of("glue").endpointFor(Region.of(REGION))` derives the partition-correct host.

examples/src/main/java/com/amazonaws/services/schemaregistry/examples/kds/PutRecordGetRecordExample.java
```diff
@@ -70,7 +72,7 @@
  */
 public class PutRecordGetRecordExample {
     private static final String AVRO_USER_SCHEMA_FILE = "src/main/resources/user.avsc";
-    private static AmazonKinesis kinesisClient;
+    private static KinesisClient kinesisClient;
```
subramp: Nit:` static kinesisClient assigned from a try-with-resources local`. After the try block the static field points at a closed client. Harmless inside main, but passing the client into putRecordsWithSchema and getRecordsWithSchema and dropping the static field is cleaner.

integration-tests/src/test/java/com/amazonaws/services/schemaregistry/integrationtests/properties/GlueSchemaRegistryConnectionProperties.java
cmakineni: Good catch. Fixed in 707ec1a: `ENDPOINT` now derives the host from the region's partition via `ServiceMetadata.of("glue").endpointFor(Region.of(REGION))`, so `cn-north-1`/`us-gov-west-1` resolve to the correct suffix. For standard regions (e.g. `us-east-2`) the value is unchanged, so existing integration runs are unaffected.

examples/src/main/java/com/amazonaws/services/schemaregistry/examples/kds/PutRecordGetRecordExample.java
cmakineni: Fixed in 809148c: dropped the static `kinesisClient` field and pass the client from the try-with-resources block into `putRecordsWithSchema`/`getRecordsWithSchema`, so nothing references a closed client after the block.

isaurab007:
Ran csharp tests offline and they are all successful. Approving the PR.

isaurab007 (approved):
LGTM - ran tests for csharp offline as there is a multilang pom change also

subramp (approved):

PR #542, opened 2026-09-10
build: drop compile-scope kotlinx-serialization-core pin, upgrade Wire to 7.0.1 (fixes #313)

### LINKED ISSUE #313: Dependency conflict prevents usage of this library if a project has newer kotlinx-serialization

subramp:
## Summary

Fixes #313 by removing the compile-scope `kotlinx-serialization-core-jvm` pin instead of bumping it, and upgrades Wire to 7.0.1 so the dependency arrives at the version Wire itself ships and tests against.

Supersedes #530, which bumps the same pin to 1.7.3. That bump does fix the reported symptom for some consumers, but it keeps the declaration at compile scope, which is the actual mechanism of the breakage.

## Problem

`serializer-deserializer/pom.xml` declared `kotlinx-serialization-core-jvm` at **compile** scope. Compile scope puts the jar on every downstream consumer's `compileClasspath`, where the Kotlin serialization compiler plugin reads `Require-Kotlin-Version` from the jar manifest and hard-fails when the consumer's Kotlin compiler is older:

```
Your current Kotlin version is 1.9.25, while kotlinx.serialization core runtime
1.7.3 requires at least Kotlin 2.0.0-RC1.
```

Every `kotlinx-serialization-core` release from 1.7.0 up declares `Require-Kotlin-Version: 2.0.0-RC1`, so any bump into that range trades one set of broken consumers for another. Verified from the published jars:

| core version | `Require-Kotlin-Version` | Kotlin metadata `mv` |
|---|---|---|
| 1.4.0 (current) | 1.4.30-M1 | 1.6.0 |
| 1.6.3 | 1.4.30-M1 | 1.9.0 |
| 1.7.3 (#530) | 2.0.0-RC1 | 2.0.0 |
| 1.11.0 | 2.0.0-RC1 | 2.3.0 |

The pin also *downgraded* what Wire ships. `wire-compiler:5.2.0` asks for core 1.7.3 at runtime scope, but the explicit 1.4.0 declaration won nearest-wins, so `kaml-jvm:0.67.0` was running against a core version it was not built against. The same pin held `kotlin-stdlib` at 1.9.25 while Wire 5.2.0 asked for 2.0.21.

## Fix

GSR has zero source references to `kotlinx-serialization-core`; it is needed only at runtime by Wire's kaml dependency. Removing the declaration lets `wire-compiler` supply it transitively at runtime scope, so it never reaches a consumer's compile classpath at any version.

Resolved tree after the change:

```
+- org.jetbrains.kotlin:kotlin-stdlib:jar:2.3.20:compile
\- com.squareup.wire:wire-compiler:jar:7.0.1:compile
   +- org.jetbrains.kotlinx:kotlinx-serialization-core-jvm:jar:1.11.0:runtime
   \- com.charleskorn.kaml:kaml-jvm:jar:0.104.0:runtime
```

Changes:

- Remove the explicit `kotlinx-serialization-core-jvm` 1.4.0 declaration.
- `square-wireschema.version` 5.2.0 -> 7.0.1.
- `kotlin.version` 1.9.25 -> 2.3.20, matching the stdlib these artifacts are built against.
- `okio` and `okio-fakefilesystem` -> 3.18.2, matching Wire 7.0.1.
- Adapt `FileDescriptorUtils` to the `weakImports` parameter added to `ProtoFileElement` in Wire 7. This is the only Wire 7 API break across every Wire symbol GSR uses.

## Compatibility

Wire 7.0.1's `wire-schema-jvm`, `wire-runtime-jvm`, and `wire-compiler` are all still Java 8 bytecode, so `<source>1.8</source>`/`<target>1.8</target>` is unchanged.

The `kotlin-stdlib` bump does raise the floor for Kotlin consumers compiling against GSR, which is why this targets the 2.0.0 major rather than a patch release.

## Testing

- `mvn -pl serializer-deserializer test` — **1346/1346 pass**, 0 failures, 0 errors, including `FileDescriptorUtilsTest` and the protobuf serializer suites that cover the changed call path.
- `mvn --file integration-tests/pom.xml verify -Psurefire` against real AWS Glue Schema Registry — **74/74 pass**, BUILD SUCCESS:
  - `GlueSchemaRegistryKafkaIntegrationTest` 45/45
  - `GlueSchemaRegistryKinesisIntegrationTest` 28/28
  - `ProtobufSchemaConverterIntegrationTest` 1/1

- `dotnet test` on the C# multilang client — **173/173 pass**, 0 failures, 4m42s, against real AWS Glue Schema Registry in us-east-1. This is an end-to-end check of this change: the C# SerDe wraps the C layer, which wraps the GraalVM-compiled Java SerDe, and `multilang-schema-registry` depends on `schema-registry-serde` with `ProtobufPreprocessor` calling the same Wire printer path touched here. The 140 MB `libnativeschemaregistry.so` was rebuilt from this branch, with `wire-schema-jvm-7.0.1`, `kotlinx-serialization-core-jvm-1.11.0`, `kaml-jvm-0.104.0`, `okio-jvm-3.18.2` and `kotlin-stdlib-2.3.20` on the native-image classpath.

Environment for the C# run: GraalVM CE 21.0.2, CMake 3.28.3, clang-tidy 11.1.0, .NET SDK 8.0.425, Amazon Linux 2.

Not run: the Kafka Connect half of `run-local-tests.sh` (MongoDB source and file sink), the C `ctest` suite (its CMake coverage setup requires `lcov`, which has no Amazon Linux 2 package), and the remaining Java modules were not built.

Separately, note that `integration-tests/docker-compose.yml` cannot start as written today, because `public.ecr.aws/bitnami/kafka` and `public.ecr.aws/bitnami/zookeeper` no longer exist upstream. I worked around it locally to run the integ tests; it is not part of this change and deserves its own fix.

Two further pre-existing issues surfaced while running the C# suite, both unrelated to this change (`git diff 04dc718 HEAD -- multilang-schema-registry` is empty). Detail is in the comment below: the first `dotnet build` on a clean clone always fails because `<Compile Include=".proto/*.cs" />` is globbed before Grpc.Tools generates those files, and `multilang-schema-registry/README.md` documents a build order that does not work (the profile is `native-image-glibc`, and CMake, not Maven, is the entry point).


subramp:
## C# (multilang) verification: 173/173 pass

The C# SerDe binds to the C layer, which wraps the GraalVM-compiled Java SerDe, so it exercises this change end to end. `multilang-schema-registry` depends on `schema-registry-serde`, and `ProtobufPreprocessor` calls the same Wire printer path this PR touches. Rebuilt the native image from this branch and ran the suite against real AWS Glue Schema Registry:

```
Passed!  -  Failed: 0, Passed: 173, Skipped: 0, Total: 173, Duration: 4 m 42 s - AWSGsrSerDe.Tests.dll (net8.0)
```

Confirmed the shared library was built with this PR's dependency set, not a cached one:

```
libnativeschemaregistry.so   140 MB
  wire-schema-jvm-7.0.1.jar
  kotlinx-serialization-core-jvm-1.11.0.jar
  kaml-jvm-0.104.0.jar
  okio-jvm-3.18.2.jar
  kotlin-stdlib-2.3.20.jar
```

Toolchain: GraalVM CE 21.0.2, CMake 3.28.3, clang-tidy 11.1.0, .NET SDK 8.0.425, Amazon Linux 2. Region us-east-1.

### Notes for maintainers, all pre-existing and unrelated to this PR

`multilang-schema-registry/csharp` is byte-identical between this branch and master (`git diff 04dc718 HEAD -- multilang-schema-registry` is empty), so none of the friction below is caused by this change. Recording it because it cost time and is worth separate fixes:

1. **First `dotnet build` on a clean clone always fails.** Two `CS0234` errors for `Basicsyntax3.Dollar` and `Basicsyntax2.bar`. `AWSGsrSerDe.Tests.csproj` declares `<Compile Include=".proto/*.cs" />`, but MSBuild evaluates that glob at project load, before `Grpc.Tools` runs protoc into `.proto/`. The files are generated correctly; they are simply not in the compile set on the first pass. A second `dotnet build` succeeds. A `Target`-ordered include or an explicit `AddCompileItems` after codegen would fix it.
2. **`cmake -S. -Bbuild` aborts without `lcov`**, with `lcov not found! Aborting...` from `test/cmake/CodeCoverage.cmake:130`. `lcov` is not packaged for Amazon Linux 2 (`No package lcov available`), even though the module README lists AL2 as a target. I configured with `-DDISABLE_TESTS=ON -DDISABLE_QUALITY_GATES=ON` to get past it, which means the C `ctest` suite did not run here. `src/CMakeLists.txt` also sets `CMAKE_C_CLANG_TIDY` unconditionally on Linux, so `clang-tidy` is a hard build requirement.
3. **The README build order does not work.** It says `mvn install -P native-image`, but the profile is `native-image-glibc`, and running Maven first fails with `Error: Invalid library path .../c/build/src`. The real entry point is the CMake build, which invokes the Maven native-image profile as a `POST_BUILD` step. Worth correcting in `multilang-schema-registry/README.md`.
4. **NuGet restore fails on Amazon Linux 2** with `NU3018` for a dozen transitive packages, because the .NET 8 signature trust store is not populated there. Worked around with `DOTNET_NUGET_SIGNATURE_VERIFICATION=false`.

Java-side results for this PR remain 1346/1346 unit and 74/74 integration, as in the description.

knoxy5467 (approved):

PR #537, opened 2026-09-01
feat(common): make the HTTP client injectable in AWSSchemaRegistryClient

cmakineni:
## What & why

`AWSSchemaRegistryClient` hard-codes `UrlConnectionHttpClient` for the internal `GlueClient`, which triggers "Multiple HTTP implementations were found on the classpath" (#120) and breaks IRSA/STS credential resolution in EKS (#69, #151, #288).

This adds an **opt-in** HTTP client override, leaving the default (`UrlConnectionHttpClient`) unchanged.

## Change (additive)

- New optional API `GlueSchemaRegistryConfiguration.get/setHttpClientBuilder(SdkHttpClient.Builder<?>)`, default `null` → falls back to `UrlConnectionHttpClient`:
  ```java
  config.setHttpClientBuilder(ApacheHttpClient.builder());
  ```
- Not read from the property map (no new config key). Wire-format impact: **none**. New runtime dependency: **none** (`apache-client` is test-scope).
- When a builder is injected, `proxyUrl` is ignored (caller owns proxy on their builder) — logs a WARN if both are set. Default path applies `proxyUrl` as before.

## Backwards compatibility

Behavior-contract checklist: no boxes checked — backwards-compatible, no version change in this PR. `CrossClientWireCompatIntegrationTest` proves it against real Glue: the injected path produces **byte-identical** output to the default path, so old→new and new→old reads are unchanged in all combinations. C# never sets this Java-only option, so it's unaffected.

## Also in this PR: integration-test infra fix

`integration-tests/docker-compose.yml` pinned `public.ecr.aws/bitnami/kafka:2.8`, which no longer exists — the tests couldn't run. Updated to a single-node **KRaft** `apache/kafka:3.8.0`.

## Tests

- **Unit (`common`):** 4 new tests (injected builder used; Apache client builds; default path unchanged; proxy + injected builder). `common` suite 143 passing, checkstyle clean, jacoco ≥85%.
- **Integration (real Glue, us-east-2 — not run in CI):** full `verify -Psurefire` suite **74/0** — Kinesis 28, Kafka 45, `CrossClientWireCompatIntegrationTest` 1.
- **C# (multilang):** GraalVM native image over the patched Java; `AWSGsrSerDe` **173/0**, including the serialize→register→deserialize round-trip.

PR #535, opened 2026-08-13
fix(integration-tests): tolerate ResourceInUseException in Kinesis setUp

isaurab007:
GlueSchemaRegistryKinesisIntegrationTest.setUp fails intermittently with

  ResourceInUseException: Stream gsr-integ-test-kinesis-stream-XXXX
  already exists (Status Code: 400) (SDK Attempt Count: 2)

LocalStack installs its Kinesis backend (kinesis-local) from npm lazily, on the first Kinesis API call, and that install blocks the call. When it is slow enough, the SDK classifies the outstanding CreateStream as a retryable failure and issues a second attempt. CreateStream is not idempotent and has no idempotency token, so the retry is a genuinely new create and is answered ResourceInUseException -- even though the first attempt had already succeeded server-side. setUp treated that as fatal.

Reference for the exception: https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/kinesis/model/ResourceInUseException.html

The postcondition setUp needs is that the stream exists, and ResourceInUseException is evidence that it does. Treat it as success and let the existing Awaitility barrier continue to gate on StreamStatus .ACTIVE. Any other cause is rethrown unchanged.

This is correct against real Kinesis as well as LocalStack: the real service answers a duplicate CreateStream the same way.

Observed on the AWS-internal GSR Java client canary, which runs this module every 20 minutes: ~3% of runs failed this way over three weeks after npm install latency for kinesis-local stepped from ~2s to ~5s (tail to 12s) on 2026-07-20.

*Issue #, if available:*
N/A - it fixes the intermittent java integ test failures

*Test results*
```
11 Aug 2026 21:57:53,551 [INFO] (main) GlueSchemaRegistryKinesisIntegrationTest: Creating Kinesis Stream : gsr-integ-test-kinesis-stream-rKHV with 1 shards on localStack..
11 Aug 2026 21:57:54,515 [INFO] (main) GlueSchemaRegistryKinesisIntegrationTest: Finished creating Kinesis Stream : gsr-integ-test-kinesis-stream-rKHV
11 Aug 2026 21:58:16,174 [INFO] (main) GlueSchemaRegistryKinesisIntegrationTest: Creating Kinesis Stream : gsr-integ-test-kinesis-stream-TeVJ with 1 shards on localStack..
[...]

PR #521, opened 2026-05-22
fix(integration-tests): pin localstack to 4.12 to make Kinesis integration tests functional again

isaurab007:
The integration test suite regressed because docker-compose.yml pulled public.ecr.aws/localstack/localstack:latest, and the recent published revision under :latest needs a pro license to work.

So, we've decided to pin to the december release [4.12](https://blog.localstack.cloud/localstack-for-aws-release-v-4-12-0/), the last image revision known to work without such restrictions.

## Before
Failed with
```
Reason: No credentials were found in the environment. Please make sure to either set the LOCALSTACK_AUTH_TOKEN variable to a valid auth token. If you are using the CLI, you can also run `localstack auth set-token`.

```

## After
Verified:
- Unit tests: BUILD SUCCESS, 9 modules, 1m 20s
- Integration: BUILD SUCCESS, 73 tests, 0 failures, 0 errors
  - Kafka: 45 tests, 12m 21s
  - Kinesis (KPL/KCL vs LocalStack 4.12): 28 tests, 8m 23s

```
[INFO] Tests run: 45, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 741.817 s - in com.amazonaws.services.schemaregistry.integrationtests.kafka.GlueSchemaRegistryKafkaIntegrationTest
[INFO] Tests run: 28, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 502.983 s - in com.amazonaws.services.schemaregistry.integrationtests.kinesis.GlueSchemaRegistryKinesisIntegrationTest
[INFO] Tests run: 73, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

### We will take an action item to migrate to a more permanent solution because, pinning is not a long term solution as mentioned in: https://blog.localstack.cloud/localstack-single-image-next-steps/

As per the blog, pinning:
1. Increases parity drift over time
2. You’ll accumulate security and compliance debt
3. CI is typically where pinning becomes fragile first
4. We’ll upgrade later” becomes a larger migration event

PR #544, opened 2026-09-14
2.0.0: document all shipping changes and complete the Scala 2.13 move

### LINKED ISSUE #313: Dependency conflict prevents usage of this library if a project has newer kotlinx-serialization

subramp:
Two commits. The first fixes the 2.0.0 CHANGELOG entry, the second completes the Scala 2.13 move that #525 intended.

## 1. `CHANGELOG.md` documented 1 of 8 shipping changes

`CHANGELOG.md` is the migration guide for 2.0.0, and the `Release 2.0.0` section documented only the JSON `className` change. Missing: #514, #532, #537, #538, #539 and #542.

Three of those omissions matter to a consumer planning an upgrade:

- **#538** removes `aws-java-sdk-sts` from the published poms, so consumers still on AWS SDK v1 must declare it themselves.
- **#539** raises `kafka-clients` to 3.9.2 at compile scope, so anyone pinning 3.6.x gets a nearest-wins conflict to resolve.
- **#542** raises the propagated Kotlin floor from 1.9.25 to 2.3.20, which drops Kotlin 1.9.x consumers.

The section now covers all eight changes, with **four breaking changes called out** (`className`, SDK v1 removal, Kotlin floor, Scala 2.12 drop) rather than one. #542 is split across three bullets, because one PR raises the Kotlin floor (breaking), removes the kotlinx pin that broke old-Kotlin builds (fix, resolves #313), and remediates a High CVE (security). The #532 entry states its remaining limitation, since #289 was auto-closed as completed while `SchemaBuilder.struct()` without `.name(...)` still throws. Test-only and demo-only changes (#520, #521, #535, #541) are omitted, matching how earlier entries treat test changes.

## 2. #525 announced a Scala 2.12 drop that was not in the build

#525 merged as `c43dc20` with the title "Move to Scala 2.13 (BREAKING: drops 2.12)", but the property bump was not in the diff. On `master`, `kafka.scala.version` was still `2.12`, so `kafka_${kafka.scala.version}` and `mbknor-jackson-jsonschema_${kafka.scala.version}` both still resolved to `_2.12`. The only pom change was a literal-to-property swap.

It looks like the bump was lost rather than omitted: commits `f7426d5` and `914f34a` on the #525 branch both set the property to 2.13, and the subsequent merge `578b3a0` ("Merge branch 'master' into fix/350-bump-kafka-scala-2.13") resolved the pom conflict back to 2.12. The squash then carried only the surviving files.

`master` was therefore left claiming a breaking change it had not made. This PR **completes the change** rather than withdrawing the claim: `kafka.scala.version` is now `2.13`, and the Scala entry is restored inside the single `Release 2.0.0` section rather than as the duplicate top-of-file heading #525 added above `Release 1.0.0`.

## Verification

Run on this branch with **JDK 8**, the JDK the release script requires:

| Check | Result |
|---|---|
| `mvn dependency:tree`, all 13 modules | `kafka_2.13:3.9.2`, `mbknor-jackson-jsonschema_2.13:1.0.39`, `scala-library` 2.13.x, **zero** remaining `_2.12` references |
| `mvn clean install -DskipTests` | **BUILD SUCCESS**, all 13 modules |
| `mvn test` on `common`, `serializer-deserializer`, `kafkastreams-serde`, and the three kafkaconnect converters | **1959 tests, 1 failure** |

`serializer-deserializer` (1346 tests) and `jsonschema-kafkaconnect-converter` (329 tests), the two modules that actually consume the Scala-suffixed `mbknor-jackson-jsonschema`, are fully green.

The single failure is `GlueSchemaRegistryConfigurationTest.testBuildConfig_noRegionConfigsSupplied_throwsException`, and it is **pre-existing and unrelated to Scala**: it fails identically with `kafka.scala.version` at `2.12`, because a region resolves from the ambient environment so the expected exception is never thrown. Worth a separate fix to make that test environment-independent.

**Not covered here:** integration tests, which need Docker and localstack. Please run `AWSKinesisEcoJavaDeser` against this branch before merging.

## Note for reviewers

This is the second attempt at the Scala 2.13 move. The first, `5737442` via #310, was reverted by #314 on 2023-12-15, with the stated reason **"This can potentially break the builds, reverting"** and no build failure recorded. The build and unit-test evidence above is meant to settle that specific question. Since Scala 2.12 and 2.13 are not binary compatible and GSR's own coordinates carry no Scala suffix, consumers on 2.12 must migrate; #350, which asked for this, is already closed, and Kafka 4.0 drops Scala 2.12 regardless.

#315, the artifactId-rename half of the cross-publish approach, remains open and is not part of this PR.

Refs #350


subramp:
## Test evidence for the Scala 2.13 move

Ran the full test surface against `e586f89`, including the integration suites that CI does not run. Summary first, detail below.

| Suite | Result |
|---|---|
| Dependency resolution, all 13 modules | `kafka_2.13:3.9.2`, `mbknor-jackson-jsonschema_2.13:1.0.39`, `scala-library` 2.13.x, **zero** remaining `_2.12` references |
| `mvn clean install -DskipTests` | **BUILD SUCCESS**, 13/13 modules |
| Java unit tests | **1959 tests, 1 failure** (pre-existing, see below) |
| **Java integration tests** | **78 tests, 0 failures, 0 errors**, BUILD SUCCESS in 20:25 |
| C# unit tests (the set CI runs) | **88/88 passed** |
| **C# integration tests** | **non-deterministic**, see below. Not a signal about this change |
| GitHub Actions on `e586f89` | 15/15 green, including JDK 8/11/17 on ubuntu and windows, `build-musl`, and 4 `build-csharp-multi-distro` jobs |

### Java integration tests

Run with `mvn -Psurefire verify -pl integration-tests`, JDK 8, against real infrastructure: a KRaft `apache/kafka:3.8.0` broker (the image #537 introduced), localstack 4.12 for Kinesis, and a real AWS Glue registry.

| Test class | Tests | Result | Time |
|---|---|---|---|
| `CrossClientWireCompatIntegrationTest` | 4 | pass | 9.6s |
| `GlueSchemaRegistryKafkaIntegrationTest` | 45 | pass | 715s |
| `ProtobufSchemaConverterIntegrationTest` | 1 | pass | 0.7s |
| `GlueSchemaRegistryKinesisIntegrationTest` | 28 | pass | 495s |

Logs show real round-trips for Avro, JSON (`JsonDataWithSchema`) and Protobuf through `kafka-clients` 3.9.2 with `kafka_2.13` on the classpath.

### The one Java unit-test failure is pre-existing

`GlueSchemaRegistryConfigurationTest.testBuildConfig_noRegionConfigsSupplied_throwsException` fails, expecting an `AWSSchemaRegistryException` that is never thrown. It is **not caused by this change**: with the property reverted to `2.12` in the same environment it fails identically. The cause is environmental, a region resolving from the ambient AWS configuration, so the "no region supplied" precondition does not hold. It would be worth making that test independent of ambient configuration, separately from this PR.

### C# was tested against a freshly built native image

Since the C# binding calls into a GraalVM native image of the Java serde, testing it meaningfully required rebuilding that image from this branch rather than reusing an older `.so`:

- GraalVM CE **21.0.2**, matching `build-csharp.yml`'s `java-version: 21.0.2`
- `mvn -U clean install -Dcheckstyle.skip=true -DskipTests` under GraalVM, then the cmake C build, producing a fresh 140 MB `libnativeschemaregistry.so` plus the C libraries
- `dotnet` 8.0.425

C# unit tests: **88/88 passed**.

### C# integration tests are non-deterministic, independent of this change

Three consecutive runs of the **identical binary**:

| Run | Passed | Failed |
|---|---|---|
| 1 | 19 | 4 |
| 2 | 22 | 1 |
| 3 | 12 | 11 |

Every failure has the same cause:

```
Registry status is: DELETING. RegistryArn: arn:aws:glue:us-east-1:<redacted>:registry/native-test-registry
```

`AWSGsrSerDe.Tests/utils/IntegrationTestBase.cs` shares one registry, `native-test-registry` (`CUSTOM_REGISTRY_NAME`), across the suite and deletes it during cleanup. Glue registry deletion is asynchronous, so any test creating a schema while a previous delete is still settling fails. Same binary plus different timing equals different results, which rules out a code cause.

In its current form this suite cannot serve as a release gate. Two options: give each run a unique registry name, which is what the Java integration tests already do with UUID and timestamp suffixes, or wait for deletion to complete in teardown. Happy to file that separately if useful.

### Not covered

- C# **integration** tests remain effectively uncovered, for the reason above rather than because of this change.
- These runs were on Linux/x64 only. Cross-platform coverage comes from the CI matrix.

subramp (commented):
Updating kafka.scala.version to 2.13

isaurab007 (approved):

PR #545, opened 2026-09-15
build(msk-iam): produce a javadoc jar so Maven Central accepts the bundle

subramp:
## Problem

Publishing 2.0.0 to Maven Central failed validation:

```
pkg:maven/software.amazon.glue/schema-registry-serde-msk-iam@2.0.0:
 - Javadocs must be provided but not found in entries
```

The build itself succeeded and every artifact was signed; the Central Portal rejected the bundle, so nothing was published.

## Cause

The parent pom sets:

```xml
<excludePackageNames>*.avrodata, *.apicurio, com.amazonaws.services.schemaregistry</excludePackageNames>
```

The only main source in this module is `src/main/java/com/amazonaws/services/schemaregistry/Dummy.java`, which is in exactly that excluded package. So javadoc has nothing to document, `maven-javadoc-plugin` attaches no artifact, **exits successfully and prints no warning**, and validation fails later at the portal.

Reproduced on `v2.0.0`: `mvn javadoc:jar -pl serializer-deserializer-msk-iam` exits 0 and leaves only `target/javadoc-bundle-options`, no jar.

## Fix

Override `excludePackageNames` for this module alone. `mvn javadoc:jar` now produces `schema-registry-serde-msk-iam-2.0.0-javadoc.jar`.

## Scope check

A reactor-wide `mvn javadoc:jar` confirms every other published module already produces a javadoc jar:

| Module | javadoc jar |
|---|---|
| common, serializer-deserializer, **serializer-deserializer-msk-iam**, kafkastreams-serde, avro-kafkaconnect-converter, avro-flink-serde, multilang-schema-registry, examples, jsonschema-kafkaconnect-converter, protobuf-kafkaconnect-converter | yes |
| build-tools, integration-tests | no, and expected |

`build-tools` and `integration-tests` have **zero** main sources, so javadoc cannot produce anything for them. Their 1.1.27 artifacts on Central carry no javadoc jar either, so their shape is unchanged from a release that was accepted.

## Worth knowing

1.1.27 **did** publish a javadoc jar for this module under the same configuration, and nothing in this module changed between 1.1.27 and 2.0.0 except the version string. So the javadoc behaviour differs by toolchain or JDK build, and previously an empty jar was emitted where now nothing is. This change makes the outcome explicit instead of depending on that.

Follow-up worth considering separately: the parent's blanket `com.amazonaws.services.schemaregistry` exclusion means the primary package is undocumented across the project, which is probably not the intent.


subramp:
## Test evidence

Ran the full suite against this branch, including the integration tests CI does not run. Native image rebuilt from this branch so the C# layer exercised this code rather than a cached `.so`.

| Suite | Result |
|---|---|
| `mvn clean install -DskipTests`, 13 modules | **BUILD SUCCESS** |
| **Java integration tests** | **78 tests, 0 failures, 0 errors**, BUILD SUCCESS in 20:28 |
| C# unit tests | **88/88 passed** |
| C# integration tests | **23/23 on a clean run**, flaky on repeats (see below) |
| `mvn javadoc:jar`, target module | now produces `schema-registry-serde-msk-iam-2.0.0-javadoc.jar` (24 KB); previously produced **no jar at all** |
| `mvn javadoc:jar`, whole reactor | every other published module already produced one; no further gaps |

### Java integration detail

Real infrastructure: KRaft `apache/kafka:3.8.0`, localstack 4.12 for Kinesis, real AWS Glue.

| Class | Tests | Result |
|---|---|---|
| `CrossClientWireCompatIntegrationTest` | 4 | pass |
| `GlueSchemaRegistryKafkaIntegrationTest` | 45 | pass, 725s |
| `ProtobufSchemaConverterIntegrationTest` | 1 | pass |
| `GlueSchemaRegistryKinesisIntegrationTest` | 28 | pass, 488s |

### C# integration flakiness is pre-existing, not from this change

Three consecutive runs of the same binary: **23 passed / 0 failed**, then 22/1, then 18/5. Every failure is `Registry status is: DELETING` against the shared `native-test-registry`.

`AWSGsrSerDe.Tests/utils/IntegrationTestBase.cs` shares one registry across the suite and deletes it during cleanup. Glue registry deletion is asynchronous, so a run starting soon after a previous one fails while the delete settles. The first run after a quiet period passes cleanly. Same behaviour was observed on #544, so it is independent of both changes.

Suggested fix, separately: give each run a unique registry name, as the Java integration tests already do with UUID and timestamp suffixes.

### Scope note

This PR is 12 added lines in one pom, configuring `maven-javadoc-plugin`. It cannot alter runtime behaviour, and the Java and C# numbers above match what #544 produced on the same tree. The result that actually exercises this change is the `javadoc:jar` row: no jar before, a 24 KB jar after.

isaurab007 (approved):

